PR #15, opened 2023-07-31
feat: throw error on invalid syntax

### LINKED ISSUE #9: Return error when any issue is encountered when parsing
Right now ANTLR does soft fails (it prints the errors to STDERR, but still attempts to parse)

Screenshot 2023-07-27 at 12 08 54 PM

rhamzeh:
## References

- closes #9 

### Sources
- https://www.antlr.org/api/Java/org/antlr/v4/runtime/ANTLRErrorListener.html
- https://stackoverflow.com/questions/18132078/handling-errors-in-antlr4?rq=4
- https://stackoverflow.com/questions/20828864/accumulating-collecting-errors-via-errorlistener-to-handle-after-the-parse

## Review Checklist

### DIFF
--- a/pkg/go/Makefile
+++ b/pkg/go/Makefile
@@ -10,7 +10,7 @@ clean:
 	go clean
 
 test:
-	go test ./...
+	go test ./... -count=1
 
 lint:
 	go install github.com/golangci/golangci-lint/cmd/golangci-lint@latest && golangci-lint run
--- a/pkg/go/transformer/dsltojson_test.go
+++ b/pkg/go/transformer/dsltojson_test.go
@@ -1,7 +1,6 @@
 package transformer_test
 
 import (
-	"log"
 	"testing"
 
 	pb "buf.build/gen/go/openfga/api/protocolbuffers/go/openfga/v1"
@@ -10,10 +9,10 @@ import (
 	"google.golang.org/protobuf/encoding/protojson"
 )
 
-func TestDSLToJSONTransformer(t *testing.T) {
-	testCases, err := LoadValidTransformerTestCases()
+func TestDSLToJSONTransformerForValidCases(t *testing.T) {
+	testCases, err := loadValidTransformerTestCases()
 	if err != nil {
-		log.Fatal(err)
+		t.Fatal(err)
 	}
 
 	for _, testCase := range testCases {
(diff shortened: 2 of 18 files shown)


pkg/go/Makefile
```diff
@@ -10,7 +10,7 @@ clean:
 	go clean
 
 test:
-	go test ./...
+	go test ./... -count=1
```
adriantam, line 13: why do you need count=1?

pkg/go/transformer/dsltojson_test.go
```diff
@@ -37,4 +39,21 @@ func TestDSLToJSONTransformer(t *testing.T) {
 			require.JSONEq(t, string(jsonBytes), string(bytes))
 		})
 	}
+
+	testCases2, err := LoadInvalidDslSyntaxTestCases()
```
adriantam, line 43: Will it make sense to break it up into two different function?

pkg/go/transformer/dsltojson_test.go
```diff
@@ -11,7 +11,7 @@ import (
 )
 
 func TestDSLToJSONTransformer(t *testing.T) {
-	testCases, err := LoadTransformerTestCases()
+	testCases, err := LoadValidTransformerTestCases()
 	if err != nil {
 		log.Fatal(err)
```
adriantam, line 16: Since this is test, why not t.Fatal(err)?

pkg/go/transformer/dsltojson_test.go
```diff
@@ -37,4 +39,21 @@ func TestDSLToJSONTransformer(t *testing.T) {
 			require.JSONEq(t, string(jsonBytes), string(bytes))
 		})
 	}
+
+	testCases2, err := LoadInvalidDslSyntaxTestCases()
+	if err != nil {
+		log.Fatal(err)
```
adriantam, line 45: instead of log.Fatal(err), use t.Fatal(err)

pkg/go/transformer/jsontodsl_test.go
```diff
@@ -9,7 +9,7 @@ import (
 )
 
 func TestJSONToDSLTransformer(t *testing.T) {
-	testCases, err := LoadTransformerTestCases()
+	testCases, err := LoadValidTransformerTestCases()
 	if err != nil {
 		log.Fatal(err)
```
adriantam, line 14: use t.Fatal if possible

pkg/go/transformer/testcases_test.go
```diff
@@ -1,34 +1,31 @@
...
 	"os"
 	"path/filepath"
 )
 
-type TestCases struct {
-	Cases []TestCase
-}
-
-type TestCase struct {
+type ValidTestCase struct {
```
adriantam, line 9: Are we exporting these?  If not, why capital on struct and function names?

pkg/go/Makefile
rhamzeh: To disable the cache, otherwise rerunning tests after modifying the test data would only run the previous cached tests

PR #34, opened 2023-08-16
feat: Adding validateDsl and adapt tests

d-jeffery:
## Description

* Refactored `transformDslToJSON` to share code
* Updated test suite for additional expect verification with expected test case content
* Updated test data to list out an array of expected errors from lexer / parser.

## References

* Addresses https://github.com/openfga/language/issues/30 for existing, working cases

## Review Checklist

### DIFF
--- a/pkg/js/transformer/dsltojson.test.ts
+++ b/pkg/js/transformer/dsltojson.test.ts
@@ -3,7 +3,7 @@ import {
   loadValidTransformerTestCases, 
   loadDslSyntaxErrorTestCases
 } from "./_testcases";
-import transformDslToJSON from "./dsltojson";
+import transformDslToJSON, { validateDsl } from "./dsltojson";
 
 describe("dslToJSON", () => {
   const testCases = loadValidTransformerTestCases();
@@ -30,13 +30,24 @@ describe("dslToJSON", () => {
 
   const testCase3 = loadDslSyntaxErrorTestCases();
   testCase3.forEach((testCase) => {
-    const errorsCount = testCase.expectedError.length;
+
+    const errorsCount = testCase.expected_errors.length;
     it(`case ${testCase.name} should throw ${errorsCount} errors`, () => {
 
-      if (errorsCount === 0) {
-        expect(() => transformDslToJSON(testCase.dsl)).not.toThrow();
-      } else {
-        expect(() => transformDslToJSON(testCase.dsl)).toThrow(testCase.error_message);
+      const result = validateDsl(testCase.dsl);
+
+      expect(result.errors.length).toEqual(errorsCount);
+
+      if (result.errors.length) {
+        expect(result.message).toEqual(testCase.error_message);
+
+        for(let i = 0; i < result.errors.length; i++) {
+          const expectedError = testCase.expected_errors[i];
+
+          expect(result.errors[i].msg).toEqual(expectedError.msg);
+          expect(result.errors[i].line).toEqual(expectedError.line);
+          expect(result.errors[i].column).toEqual(expectedError.column);
+        }
       }
     });
   });
--- a/pkg/js/transformer/dsltojson.test.ts
+++ b/pkg/js/transformer/dsltojson.test.ts
@@ -1,9 +1,8 @@
 import { 
   loadInvalidDslSyntaxTestCases,
-  loadValidTransformerTestCases, 
-  loadDslSyntaxErrorTestCases
+  loadValidTransformerTestCases,
 } from "./_testcases";
-import transformDslToJSON, { validateDsl } from "./dsltojson";
+import transformDslToJSON from "./dsltojson";
 
 describe("dslToJSON", () => {
   const testCases = loadValidTransformerTestCases();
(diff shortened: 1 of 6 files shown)


tests/data/dsl-syntax-validation-cases.yaml
```diff
@@ -52,16 +42,11 @@ dsl: |
...
-  - endColumn: 22
-    endLineNumber: 6
-    message: Assignable relation 'member' must have types
-    extraInformation:
-      error: assignable-relation-must-have-type
-    severity: 8
-    source: linter
-    startColumn: 20
-    startLineNumber: 6
+expected_errors:
```
rhamzeh, line 45:
```suggestion
```

pkg/js/transformer/dsltojson.ts
```diff
@@ -269,9 +264,32 @@ export default function transformDslToJSON(dsl: string): AuthorizationModel {
...
 
   return listener.authorizationModel as AuthorizationModel;
 }
+
+/**
+ * validateDSL - Validates model authored in FGA DSL syntax, returning all found errors
+ * @param {string} dsl 
+ * @returns {OpenFgaDslSyntaxMultipleError}
+ */
+export function validateDsl(dsl: string): OpenFgaDslSyntaxMultipleError {
```
rhamzeh, line 290: Move this out into `pkg/js/validator/validate-dsl.ts`

pkg/js/transformer/dsltojson.test.ts
```diff
@@ -30,13 +30,24 @@ describe("dslToJSON", () => {
 
   const testCase3 = loadDslSyntaxErrorTestCases();
   testCase3.forEach((testCase) => {
```
rhamzeh, line 32: Move these tests to `pkg/js/validator/validate-dsl.test.ts`

pkg/js/validator/validate-dsl.ts
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ * validateDSL - Validates model authored in FGA DSL syntax, returning all found errors
+ * @param {string} dsl 
+ * @returns {OpenFgaDslSyntaxMultipleError}
+ */
+export default function validateDsl(dsl: string): OpenFgaDslSyntaxMultipleError {
+    const { errorListener } = parseDSL(dsl);
+  
+    return new OpenFgaDslSyntaxMultipleError(errorListener.errors);
```
rhamzeh, line 11: I'm not sure about this - shouldn't this be throwing errors, and returning null if the model is valid?

Here you are always returning an error.

d-jeffery: Happy to talk about this one. My reason;
* If you go throw error, return null - you need to try catch to collect failures. If i return an array of errors, its just a functional call.
* On the case with no errors, `errorListener.errors` is an empty array, so if you parse correctly the function returns that empty array.

Either way will work - I just leant this way to avoid try-catching for every invocation.

adriantam: If we are staying with the return error/nil method, maybe we should change the wording of [validate-dsl.test.ts](https://github.com/openfga/language/pull/34/files#diff-44c1dbcbe9b8294cd670c05737a815d5df7dc41587993cc486a518e622928a1a) as we are not throwing error.

d-jeffery: Yep, can do that.

Do you have any strong opinions supporting either method?

rhamzeh: I definitely do not like always returning an error, it must be a nil if it is valid. You should not get an error that has an empty array for errors

rhamzeh: And I personally prefer that we throw. It makes more sense to me. When we validate, we listen to the error and show it to the user, and if an error is not thrown, we proceed normally

d-jeffery: Cool - fair reasoning, not going to argue. I'll make the change in favor of the throw.

pkg/js/validator/validate-dsl.test.ts
```diff
@@ -0,0 +1,45 @@
...
+describe("validateDsl", () => {
+  const testCases = loadDslSyntaxErrorTestCases();
+  testCases.forEach((testCase) => {
+
+    const errorsCount = testCase.expected_errors.length;
+    it(`case ${testCase.name} should throw ${errorsCount} errors on validation`, () => {
+
+      try {
+        validateDsl(testCase.dsl);
+      } catch(thrownError) {
```
rhamzeh, line 14:
```suggestion
      } catch(thrownError as OpenFgaDslSyntaxMultipleError) {
```

Can't you do this instead of copying it into an exception below?

If it doesn't work:

```ts
thrownError as any as OpenFgaDslSyntaxMultipleError
```

pkg/js/validator/validate-dsl.test.ts
```diff
@@ -0,0 +1,45 @@
...
+            expect(exception.errors[i].column).toEqual(expectedError.column);
+            
+            if (expectedError.metadata) {
+              const resultMetadata = exception.errors[i].metadata;
+              const expectedMetadata = expectedError.metadata;
+  
+              expect(resultMetadata?.symbol).toEqual(expectedMetadata.symbol);
+              expect(resultMetadata?.start).toEqual(expectedMetadata.start);
+              expect(resultMetadata?.stop).toEqual(expectedMetadata.stop);
+            }
```
rhamzeh, line 37:
```suggestion
            } else {
              expect(resultMetadata).toBeUndefined();
            }
```

pkg/js/validator/validate-dsl.test.ts
```diff
@@ -0,0 +1,45 @@
...
+  
+          for (let i = 0; i < errorsCount; i++) {
+            const expectedError = testCase.expected_errors[i];
+  
+            expect(exception.errors[i].msg).toEqual(expectedError.msg);
+            expect(exception.errors[i].line).toEqual(expectedError.line);
+            expect(exception.errors[i].column).toEqual(expectedError.column);
+            
+            if (expectedError.metadata) {
+              const resultMetadata = exception.errors[i].metadata;
```
rhamzeh, line 31: Move these two outside the if, and in the if check against:

`if (expectedMetadata) {...`

pkg/js/validator/validate-dsl.test.ts
d-jeffery: Doesn't like either.

pkg/js/validator/validate-dsl.test.ts
d-jeffery: This caught a missing metadata from one of the tessts.

pkg/js/validator/validate-dsl.test.ts
d-jeffery: Done

pkg/js/validator/validate-dsl.test.ts
rhamzeh: We'll deal with it later

rhamzeh (approved):

PR #43, opened 2023-09-07
fix: address and test misspelled define

d-jeffery:
## Description

Fix for null pointer when define was misspelled.

## Review Checklist

### DIFF
--- a/pkg/js/transformer/dsltojson.ts
+++ b/pkg/js/transformer/dsltojson.ts
@@ -88,9 +88,16 @@ class OpenFgaDslListener extends OpenFGAListener {
   };
 
   exitRelationDeclaration = (ctx: RelationDeclarationContext) => {
+    if (!ctx.relationName()) {
+      return;
+    }
+
     const relationName = ctx.relationName().getText();
     let relationDef: Userset | undefined;
     const rewrites = this.currentRelation?.rewrites;
+    if (!rewrites?.length) {
+      return;
+    }
     if (rewrites?.length === 1) {
       relationDef = rewrites[0];
     } else {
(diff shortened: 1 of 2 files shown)


pkg/js/transformer/dsltojson.ts
```diff
@@ -116,6 +115,15 @@ class OpenFgaDslListener extends OpenFGAListener {
       }
     }
     if (relationDef) {
+      const relationName = ctx.relationName().getText();
```
rhamzeh, line 118: Why not do:

```js
if (!ctx.relationName()) {
  return;
}
```

At the start?

There is no case where you have no relation name or no relation definition and where you should proceed as usual

rhamzeh (changes requested):

pkg/js/transformer/dsltojson.ts
```diff
@@ -116,6 +120,13 @@ class OpenFgaDslListener extends OpenFGAListener {
       }
     }
     if (relationDef) {
+      // Throw error if same named releation occurs more than once in a relationship definition block
```
rhamzeh, line 123:
```suggestion
      // Throw error if same named relation occurs more than once in a relationship definition block
```

pkg/js/transformer/dsltojson.ts
```diff
@@ -88,6 +88,10 @@ class OpenFgaDslListener extends OpenFGAListener {
   };
 
   exitRelationDeclaration = (ctx: RelationDeclarationContext) => {
+    if (!ctx.relationName()) {
+      return;
+    }
+
```
rhamzeh, line 94: Can you also add a check on line 98? `if (!rewrites?.length) { return }`

It should not be possible to have a relation with no definition

rhamzeh (approved):

PR #129, opened 2024-01-03
chore: update dependencies

rhamzeh:
## Description

Go:
- Set target go to 1.20
- Update dependencies

JS:
- Fix types
- Update dependencies
- Release v0.2.0-beta.8

Note: Go test are expected to fail as it does not support nesting yet

## Review Checklist


pkg/go/go.mod
```diff
@@ -1,13 +1,13 @@
 module github.com/openfga/language/pkg/go
 
-go 1.21.5
+go 1.20
```
sergiught, line 3: What's causing is to downgrade?

ewanharris (approved):

pkg/go/go.mod
rhamzeh: A third party faced an issue with us using go1.21, we're gonna do a latest - 1 approach, so support the latest minor + the one before it. So while current is go1.21, we'll support go1.20+

See here: https://github.com/lxc/incus/issues/315#issuecomment-1863381991

sergiught: Understood:) thanks for the reference. Should we then publicly follow the same release policy as Go itself and state it publicly within all our customer facing SDKs/Tools https://go.dev/doc/devel/release#policy?

rhamzeh: I dunno if we want to publicly commit to it yet versus a "best effort". At least for now

sergiught (approved):

PR #107, opened 2023-11-13
feat: add limited support for mixed operators

rhamzeh:
This PR adds initial support for mixed operators

## Description

Cases to be supported (both from and to DSL)
(with no direct assignment):
- [x] `define relation: (rel1 and rel2) but not rel3`
- [ ] `define relation: ((rel1 and rel2) but not rel3)` # transformation back to DSL will not keep the extraneous surrounding brackets
- [x] `define relation: rel1 and (rel2 but not rel3)`
- [x] `define relation: rel1 and ((rel2 but not rel3) or (rel4 and (rel5 but not rel6 from rel7)))`

(with direct assignment, only if there is a single reference to the assignment and only if it is the first entry):
- [x] `define relation: [user, user:*, user#follower with condition] or ((rel1 and rel2) but not rel3)`
- [x] `define relation: ([user] or ((rel1 and rel2)) but not rel3` # Note we can de-prioritize this as pending discussion

Cases NOT planned to supported yet, pending discussion:
Anything involving multiple references to direct assignment, or the presence of direct assignment not as the first entry, or references to `self/this`:
- `define relation: [user] as self | rel1 but not self`
- `define relation: [user] as self | this but not rel1`
- `define relation: rel1 but not [user]`
- `define relation: [user] as self | (self or ((rel1 and rel2)) but not rel3`
- `define relation: (rel3 or ((rel1 and rel2)) but not [user]`
- `define relation: (rel3 or (([user] and rel2)) but not self`
- `define relation: ([user] and allowed) or ([user] but not blocked]`

## References

- Closes #16 

## Review Checklist
- [ ] I have clicked on ["allow edits by maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [ ] I have added documentation for new/changed functionality in this PR or in a PR to [openfga.dev](https://github.com/openfga/openfga.dev) [Provide a link to any relevant PRs in the references section above]
- [ ] The correct base branch is being used, if not `main`
- [ ] I have added tests to validate that the change in functionality is working as expected

PR #136, opened 2024-01-23
feat: add limited support for mixed operators in Golang binding

### LINKED ISSUE #132: Support nested userset rewrite operands in the DSL in the Go bindings
Related to #16

Currently the Go bindings do not properly support mixed operators. Work has happened in #107 to add the bindings to JS

The following will result in incorrect JSON

Tests for these cases are already in place and the Go implementation fails on them

```py
 define relation: (rel1 and rel2) but not rel3
 define relation: ((rel1 and rel2) but not rel3) # transformation back to DSL will not keep the extraneous surrounding brackets
 define relation: rel1 and (rel2 but not rel3)
 define relation: rel1 and ((rel2 but not rel3) or (rel4 and (rel5 but not rel6 from rel7)))
 define relation: [user, user:*, user#follower with condition] or ((rel1 and rel2) but not rel3)
 define relation: ([user] or ((rel1 and rel2)) but not rel3 
```

jimmyjames:
This PR adds initial support for mixed operators to the Golang bindings.

## Description

Golang binding implementation for the initial mixed operators support as done in #107. Similar to the JS implementation, this change does not support anything involving multiple references to direct assignment, or the presence of direct assignment not as the first entry, or references to `self/this`.

## References
- Golang implementation of JS bindings in #107 
- Closes #132 

## Review Checklist

### DIFF
--- a/pkg/go/transformer/jsontodsl.go
+++ b/pkg/go/transformer/jsontodsl.go
@@ -10,6 +10,58 @@ import (
 	"google.golang.org/protobuf/encoding/protojson"
 )
 
+type DirectAssignmentValidator struct {
+	occurred   int
+	stateStack []pb.Userset
+}
+
+func (v *DirectAssignmentValidator) incr() {
+	v.occurred++
+}
+
+func (v *DirectAssignmentValidator) occurrences() int {
+	return v.occurred
+}
+
+func (v *DirectAssignmentValidator) reset() {
+	v.stateStack = []pb.Userset{}
+	v.occurred = 0
+}
+
+var validator = DirectAssignmentValidator{
+	occurred:   0,
+	stateStack: make([]pb.Userset, 0),
+}
+
+func (v *DirectAssignmentValidator) isFirstPosition(userset *pb.Userset) bool {
+	if userset.GetThis() != nil {
+		return true
+	}
+
+	// TODO nil handling
+	if userset.GetDifference().GetBase() != nil {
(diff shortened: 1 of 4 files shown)


pkg/go/transformer/jsontodsl.go
```diff
@@ -10,6 +10,59 @@ import (
...
+func (v *DirectAssignmentValidator) occurrences() int {
+	return v.occurred
+}
+
+func (v *DirectAssignmentValidator) reset() {
+	v.stateStack = []pb.Userset{}
+	v.occurred = 0
+}
+
+var validator = DirectAssignmentValidator{
```
rhamzeh, line 31: Similar to the JS implementation, I think we should not be using a global variable and rely instead on an instantiated instance that's passed around for the duration of the validation.

However - considering how excited I am to ship this feature and have all the tests passing, let's defer that to another PR.

I would love to chat w/ both you and @d-jeffery if the global variable is required for some reason that I'm not considering.

rhamzeh (approved):
Thanks @jimmyjames! ❤️ 

Can't wait to have this out!

pkg/go/transformer/jsontodsl.go
jimmyjames: @rhamzeh that makes sense to me; I wanted to start with a fairly direct implementation of the existing JS implementation to get something out the door, then I think we can refactor both to a non-global. Want me to create an issue for that?

PR #147, opened 2024-02-12
feat: Validation for JSON models

d-jeffery:
## Description

- resolves https://github.com/openfga/language/issues/100

## Review Checklist

### DIFF
--- a/pkg/js/tests/validator.test.ts
+++ b/pkg/js/tests/validator.test.ts
@@ -1,6 +1,12 @@
-import { loadDSLSyntaxErrorTestCases, loadDSLValidationErrorTestCases } from "./_testcases";
+import {
+  loadDSLSyntaxErrorTestCases,
+  loadDSLValidationErrorTestCases,
+  loadInvalidJSONSValidationTestCases,
+} from "./_testcases";
 import { DSLSyntaxError, DSLSyntaxSingleError, ModelValidationError, ModelValidationSingleError } from "../errors";
-import { validateDSL } from "../validator";
+import { validateDSL, validateJSON } from "../validator";
+import { transformDSLToJSON, transformDSLToJSONObject } from "../transformer";
+import { AuthorizationModel } from "@openfga/sdk";
 
 describe("validateDSL", () => {
   const syntacticTests = loadDSLSyntaxErrorTestCases();
@@ -59,6 +65,7 @@ describe("validateDSL", () => {
             `${errorsCount} error${errorsCount === 1 ? "" : "s"} occurred:\n\t* ${testCase.expected_errors
               .map((err: ModelValidationSingleError) => {
                 const errorType = err.metadata?.errorType || "validation";
+
                 return `${errorType} error at line=${err.line?.start}, column=${err.column?.start}: ${err.msg}`;
               })
               .join("\n\t* ")}\n\n`,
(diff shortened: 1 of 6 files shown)


tests/data/json-validation-cases.yaml
```diff
@@ -0,0 +1,3020 @@
+---
+- name: valid schema version
+  json: |
+    {
+      "schema_version": "1.1"
```
rhamzeh, line 5: use 1.3, because 1.2 will soon become valid :)

pkg/js/tests/validator.test.ts
```diff
@@ -72,3 +77,41 @@ describe("validateDSL", () => {
     });
   });
 });
+
+describe("validateJSON", () => {
+  const testCases = loadInvalidJSONSValidationTestCases();
```
rhamzeh, line 82: can we append the DSL ones after converting to these being run?

tests/data/json-validation-cases.yaml
rhamzeh: There is no need to have everything in here, as many can be written in DSL format and imported.

That has the additional benefit of not needing to write many of the tests twice.

Only json that is not supported by the DSL needs to have extra tests

rhamzeh (commented):
Any chance we can have `validateDSL` call `validateJSON` but with the dslString optional?

From a quick skim the main difference between the two is that `validateDSL` passes the index.

So validateJSON can do that if the DSL string metadata is provided to the function, otherwise it passes undefined. WDYT?

d-jeffery:
> Any chance we can have `validateDSL` call `validateJSON` but with the dslString optional?
> 
> From a quick skim the main difference between the two is that `validateDSL` passes the index.
> 
> So validateJSON can do that if the DSL string metadata is provided to the function, otherwise it passes undefined. WDYT?

Thats what I thought of first;

Problem I see is that every function under validateJSON expects the `lines: string[]` to be passed to it to generate a line number and col for exceptions. This is consumed by the error collector, and before every exception a `lineIndex` is calculated using `lines: string[]`.

It would mean making all the functions that currently expect `lines` and/or `lineIndex` to make the parameter `string[] | undefined` and/or `number | undefined`, then handle the undefined behavior.

e.g Of a typical exception generation

```typescript
function populateRelations(
  lines: string[],
  // eslint-disable-next-line @typescript-eslint/no-explicit-any
  collector: any,
  authorizationModel: AuthorizationModel,
  typeRegex: ValidationRegex,
  relationRegex: ValidationRegex,
) {
  // Looking at the types
  authorizationModel.type_definitions?.forEach((typeDef) => {
    const typeName = typeDef.type;

    if (typeName === Keyword.SELF || typeName === ReservedKeywords.THIS) {
      const lineIndex = getTypeLineNumber(typeName, lines);
      collector.raiseReservedTypeName(typeName, lineIndex);
    }

    if (!typeRegex.regex.test(typeName)) {
      const lineIndex = getTypeLineNumber(typeName, lines);
      collector.raiseInvalidName(typeName, typeRegex.rule, undefined, lineIndex);
    }
```    

This is main reason i split the code - if you'd prefer handling undefined parameters here, I can try that. I have already made line & col values optional for the collector, so that behavior is already working.

pkg/js/util/exceptions.ts
```diff
@@ -228,12 +240,27 @@ const createDuplicateRelationError = (props: BaseProps, relationName: string) =>
 
 const createDuplicateRelationshipDefinitionError = (props: BaseProps) => {
   const { errors, lines, lineIndex, symbol } = props;
+
+  console.dir(props);
```
rhamzeh, line 244: 🔪

pkg/js/validator/validate-dsl.ts
```diff
@@ -684,54 +695,59 @@ function modelValidation(
...
+      if (!relationRegex.regex.test(relationName)) {
         const typeIndex = getTypeLineNumber(typeName, lines);
         const lineIndex = getRelationLineNumber(relationName, lines, typeIndex);
-        collector.raiseInvalidName(lineIndex, relationName, relationRegex.rule, typeName);
-      } else if (encounteredRelationsInType[relationName]) {
+        collector.raiseInvalidName(relationName, relationRegex.rule, typeName, lineIndex);
+      }
+
+      // TODO: Will never be true, consider removing
+      if (encounteredRelationsInType[relationName]) {
```
rhamzeh, line 744: 🔪

pkg/js/validator/validate-dsl.ts
```diff
@@ -745,11 +761,11 @@ function populateRelations(
  * @param {ValidationOptions} options
  */
 export function validateJSON(
-  jsonString: string,
   authorizationModel: AuthorizationModel,
   options: ValidationOptions = {},
+  jsonString?: string,
```
rhamzeh, line 766: dslString

rhamzeh (approved):
All good from my end! You think we can rename the file from `validate-dsl` to `validate`?

ewanharris (approved):

PR #155, opened 2024-02-14
chore: bump go version to 1.21

ewanharris:
## Description

Now that 1.22 is out in theory we can bump this, it should fix the golangci-lint issues that we're seeing due to their minimum version being bumped

Alternatively, we could
* pin to a version of `golangci-lint` other than latest
* match how CI works for JS and run our vulncheck and linting on a newer version of Go and then run tests using all our supported versions of Go rather than pointing to the go.mod file.

Edit:

I had to move this to `1.21.3` due to this being the version set in https://github.com/openfga/api/blob/main/proto/go.mod and go was enforcing that be set rather than `1.21`

## Review Checklist

### DIFF
--- a/.github/workflows/main-go.yaml
+++ b/.github/workflows/main-go.yaml
@@ -31,7 +31,7 @@ jobs:
 
       - uses: actions/setup-go@0c52d547c9bc32b1aa3301fd7a9cb496313a4491 # v5.0.0
         with:
-          go-version: '1.22'
+          go-version: '>=1.22'
           cache-dependency-path: './pkg/go/go.sum'
           check-latest: true
 
--- a/.github/workflows/main-go.yaml
+++ b/.github/workflows/main-go.yaml
@@ -24,8 +24,28 @@ permissions:
   contents: read
 
 jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+
+      - uses: actions/setup-go@0c52d547c9bc32b1aa3301fd7a9cb496313a4491 # v5.0.0
+        with:
+          go-version: '1.22'
+          cache-dependency-path: './pkg/go/go.sum'
+          check-latest: true
+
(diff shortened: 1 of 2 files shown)


ewanharris:
New failure is why I think we should split out lint and vulncheck to run on a newer (or even latest) version of Go and then test against a supported range to avoid govulncheck flagging issues in an older version

pkg/go/go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/openfga/language/pkg/go
 
-go 1.21
+go 1.21.3
```
rhamzeh, line 3: I think `go1.21.7` is the only one that would quite down the alerts.

Alternatively you can force it in the `main-go` workflow file to use go>=1.21.7

ewanharris: I tried to set just `go 1.21` here, but it seems the newer Go tooling is a bit more strict around the value set here and (I think) it's the minimum from the list of dependencies.

I was leaning towards running lint/vulncheck with a newer version of Go so that we don't pickup irrelevant vuln warnings and then run tests against 1.21 and 1.22 to verify our supported range like in JS. What do you think?

.github/workflows/main-go.yaml
```diff
@@ -24,8 +24,28 @@ permissions:
...
 
 jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+
+      - uses: actions/setup-go@0c52d547c9bc32b1aa3301fd7a9cb496313a4491 # v5.0.0
+        with:
+          go-version: '1.22'
```
rhamzeh, line 34: Maybe we can do 
```
go-version: '>=1.21.7'
```

Or 
```
go-version: '>=1.22'
```
works - just so we only need to bump this on sec issues

ewanharris: Yeah that's a good idea, will change

rhamzeh (approved):

PR #159, opened 2024-02-15
feat: support use of module keyword

### LINKED ISSUE #150: In grammar, allow `module` keyword

ewanharris:
## Description

Adds support for the `module` keyword in the grammar, there's a couple things that I'm not 100% certain of (hence draft for feedback):

* Passing both `module` and `model` causes an error, but I don't know if we can improve the error message received - `no viable alternative at input '\nmodule'`
* Currently validation fails because when parsing the DSL we don't know the schema version to use so will error, I think we might need to allow passing the version obtained from a `go.mod` file into validate in some way? Which is probably outside the scope of this PR.
* I changed the existing validation to throw if we can't determine a schema version, otherwise when we get to the switch further down we can't blow up trying to determine line numbers, this might want to be changed to only throw in the case of a single file module.

## References

Closes #150, #151

## Review Checklist


rhamzeh:
We shouldn't be validating a partial model, only after all the pieces are collected can we validate

tests/data/dsl-semantic-validation-cases.yaml
```diff
@@ -1153,6 +1153,21 @@
         symbol: "allowed_ip"
         errorType: condition-not-used
 
+- name: supports module keyword
```
rhamzeh, line 1156: Yeah this is correctly failing, but the name is off.

ewanharris:
@rhamzeh I'm not sure I follow, are you saying that when processing modular models the intention is to construct the final model and then pass that to `validateDSL` rather than the individual model modules?

rhamzeh:
> when processing modular models the intention is to construct the final model and then pass that to `validateDSL` rather than the individual model modules?

Yes, the parser will handle validating the model syntactically, but you cannot validate it semantically, because it will need data from the other files.

E.g. user is not in that file, so validation will naturally fail until you have collected all the files in the model

ewanharris:
So should `validateDSL` throw an error along the lines of `cannot parse individual module` if passed a module DSL?

rhamzeh (dismissed):

PR #169, opened 2024-02-26
Merge main into modular models feature branch

ewanharris:
## Description

Merges the main branch into the modular models feature branch.

It then fixes an issue that arose due to an addition in the modular models work where we threw early, when realistically we probably shouldn't have, and a related issue to that where if we do not have a symbol (i.e. missing schema) then when we are calculating/finding line and column numbers the `findIndex`/`search` functions could return `-1` and we would then have 0 based line numbers, or `NaN` when we have attempted to add onto these values.

## Review Checklist


tests/data/dsl-semantic-validation-cases.yaml
```diff
@@ -746,8 +746,8 @@
         start: 7
         end: 7
       column:
-        start: 0
-        end: 10
+        start: 1
```
ewanharris, line 749: This is an example of where we were potentially returning a zero based line number instead of a 1 based line number

tests/data/dsl-semantic-validation-cases.yaml
```diff
@@ -1184,6 +1184,15 @@
         end: 1
       metadata:
         errorType: schema-version-required
+    - msg: invalid schema undefined
```
ewanharris, line 1187: These errors are now thrown when validating a modular model now that the early throw has been removed

rhamzeh (dismissed):

rhamzeh:
Some minor lint issues in go and js

rhamzeh (approved):

PR #177, opened 2024-03-04
feat: change transformModFileToJSON to return line and column information

### LINKED ISSUE #173: Issues in validation of fga.mod fields
`Schema` could alternatively be a number.

https://github.com/openfga/language/blob/6de767e56acd7dce22385e5845f3c660c2dec978/pkg/js/transformer/modules/mod-to-json.ts#L113

Also realized since this is just JSON, i have no range to use for marking which contents in `fga.mod` when generating errors. The structure is missing line and character values, such that i'd also need to parse the YAML as well to get line numbers.

https://github.com/openfga/language/blob/6de767e56acd7dce22385e5845f3c660c2dec978/pkg/js/transformer/modules/mod-to-json.ts#L171

ewanharris:
## Description

Changes the return type of the `transformModFileToJSON` to a structure that allows us to return line and column information as part of the data to allow us to improve error reporting without tying us to the underlying yaml parser that we use on each platform.

Currently only has the JS implementation and will follow up with the Go implementation.

## References

Resolves #173

## Review Checklist

### DIFF
--- a/pkg/go/transformer/mod-to-json.go
+++ b/pkg/go/transformer/mod-to-json.go
@@ -8,10 +8,22 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
+type ModFileStringProperty struct {
+	Value  string `json:"value"`
+	Line   int    `json:"line"`
+	Column int    `json:"column"`
+}
+
+type ModFileArrayProperty struct {
+	Value  []ModFileStringProperty `json:"value"`
+	Line   int                     `json:"line"`
+	Column int                     `json:"column"`
+}
+
 type ModFile struct {
-	Schema   string   `json:"schema"`
-	Module   string   `json:"module"`
-	Contents []string `json:"contents"`
+	Schema   ModFileStringProperty `json:"schema"`
+	Module   ModFileStringProperty `json:"module"`
+	Contents ModFileArrayProperty  `json:"contents"`
 }
 
 type YAMLModFile struct {
@@ -88,7 +100,11 @@ func TransformModFile(data string) (*ModFile, error) { //nolint:cyclop
 			Column: yamlModFile.Schema.Column,
 		})
 	default:
-		modFile.Schema = yamlModFile.Schema.Value
+		modFile.Schema = ModFileStringProperty{
+			Value:  yamlModFile.Schema.Value,
+			Line:   yamlModFile.Schema.Line,
+			Column: yamlModFile.Schema.Column,
+		}
 	}
 
 	switch {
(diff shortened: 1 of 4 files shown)


d-jeffery:
Source mapping is great: https://github.com/openfga/language/assets/1425457/52935de6-3c3f-46bf-b93e-882dc67205e7

Just questioning who should own validation of whether a file name is valid to avoid whack-a-mole.

pkg/go/transformer/mod-to-json.go
```diff
@@ -8,10 +8,22 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
+type ModFileStringProperty struct {
+	Value  string `json:"value"`
+	Line   int    `json:"line"`
+	Column int    `json:"column"`
```
rhamzeh, line 14: Should we line and column be:

```
Line: {
  Start: ...,
  End: ...
},
Column: {
  Start: ...,
  End: ...
},
```

ewanharris: It should, but we only have starting information using the current yaml package and no real way to try and calculate it :/

It might be worthwhile evaluating other yaml packages in the go ecosystem to improve our reporting here

rhamzeh: Oh that's a bummer.

It's OK we can deal with it later.

I was thinking of the case where you have a comment on the line and don't want it to show as a squiggly line, but that's very far down the prio list. Let's keep as is for now until we find something better

rhamzeh (approved):

rhamzeh:
> Just questioning who should own validation of whether a file name is valid to avoid whack-a-mole.

I think the frontend (so VS Code/CLI should always be the one validating the files, as reading the files in a uniform way is outside the scope of language), e.g. reading the files on web vs native vs code

ewanharris:
@rhamzeh  I think @d-jeffery was referring to the validation the language does currently [here](https://github.com/openfga/language/blob/chore/improve-fga-mod-file-reporting/pkg/js/transformer/modules/mod-to-json.ts#L195-L201) where it ensure that it ends with `.fga`, do you think we should remove that?

Myself and Daniel had discussed potentially moving the `.fga` check to be a "return file with errors" case and then changing the returned structure to be like `{ modFile, errors }`, that would allow us to potentially extend in future if we ever wanted to have warnings rather than just errors.

d-jeffery:
> @rhamzeh I think @d-jeffery was referring to the validation the language does currently [here](https://github.com/openfga/language/blob/chore/improve-fga-mod-file-reporting/pkg/js/transformer/modules/mod-to-json.ts#L195-L201) where it ensure that it ends with `.fga`, do you think we should remove that?

Yes, this is what I was referencing.

d-jeffery:
From a quick talk with @rhamzeh 

> agree on: file loading and validation should be done by the "UI" (VS Code ext, CLI, etc..) and file name validation being in language (this happens before loading)

> also agree on language returning a json blob containing errors with a level

PR #178, opened 2024-03-04
feat(js): implement transforming module to a model

ewanharris:
## Description

Implements the `transformModuleFilesToModel` that will take the files that represent a modular model and then returns a the full authorization model.

It handles copying over conditions and type definitions as well as ensuring they all include the `file` metadata that cannot be added during the parsing of the DSL, ensuring uniqueness between these and also potential issues with missing `extends` keywords.

It handles copying over type extensions to the original type extension, ensuring uniqueness in the relation names.

It uses some WIP versions of [js-sdk](https://github.com/openfga/js-sdk/tree/feat/modular-models-metadata) and [api/poto](https://github.com/openfga/api/tree/feat/137-modular-models-additions)

The tests need some improvement to ensure that all the bases are being covered.

## References

Is the JS portion of #153

## Review Checklist


rhamzeh:
I don't think these should be in there:
https://github.com/openfga/api/compare/main...feat/137-modular-models-additions#diff-4fff4cf94235ef3370ef32783dc14676cf70fc97cf96d3a3c13fe6bbcf018683R137-R144, these are strictly metadata.

But up for having that discussion

pkg/js/transformer/modules/modules-to-model.ts
```diff
@@ -0,0 +1,163 @@
...
+
+  try {
+    validateJSON(model as AuthorizationModel);
+  } catch (error) {
+    if (error instanceof ModelValidationError) {
+      errors.push(...error.errors);
+    }
+  }
+
+  if (errors.length !== 0) {
```
rhamzeh, line 158:
```suggestion
  if (errors.length) {
```

tests/data/transformer-module/00-simple-module/authorization-model.json
```diff
@@ -0,0 +1,52 @@
...
+        "module": "org",
+        "relations": {
+          "member": {
+            "directly_related_user_types": [
+              {
+                "condition": "a_check",
+                "type": "user"
+              }
+            ],
+            "module": "org",
```
rhamzeh, line 39: was thinking we don't include the metadata in relations unless they differ from the type. It would drastically reduce size from some of the bigger models. WDYT?

ewanharris: I've no real strong opinion here, I mainly added them to ensure consistency but not having them does have benefits like helping clearly show what is extension and what is not as well as some potential size difference gains, it might even reduce transformation time as we have to loop over all relations to add this specific one.

I'll remove now and if we realise we need them it's easy to add back.

ewanharris: Also just as I can't reply to `openfga/api` comment, agreed on removing from `RelationReference` I added there while I was still stumbling around figuring out the right places and just forgot to remove. Have removed from there and updated `js-sdk` accordingly

pkg/js/transformer/dsltojson.ts
```diff
@@ -321,6 +342,13 @@ class OpenFgaDslListener extends OpenFGAListener {
       expression: "",
       parameters: {},
     };
+
+    if (this.isModularModel) {
+      //@ts-expect-error
```
d-jeffery: Can we get a comment here for documentation?

d-jeffery (approved):
lgtm

pkg/js/transformer/dsltojson.ts
ewanharris: This is removed up now in the [commit](https://github.com/openfga/language/pull/178/commits/4707d4906f236a0d122b7a15af1e553a8445d4d3#diff-a3aecc6cfca045d5430fdd3f9c2aadd654954d204db5068a4cbb91f1e1716fe9L347) that moved to a js-sdk generated from the new protobuf definitions so the metadata fields now exist

pkg/js/transformer/dsltojson.ts
```diff
@@ -140,20 +145,31 @@ class OpenFgaDslListener extends OpenFGAListener {
...
-  exitTypeDef = (_ctx: TypeDefContext) => {
+  exitTypeDef = (ctx: TypeDefContext) => {
     if (!this.currentTypeDef?.type) {
       return;
     }
 
-    if (!Object.keys(this.currentTypeDef?.metadata?.relations || {}).length) {
+    if (this.isModularModel && !Object.keys(this.currentTypeDef?.metadata?.relations || {}).length) {
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      this.currentTypeDef!.metadata!.relations = undefined as any;
```
rhamzeh, line 161: We may nit need this, but you'll know after we do the Go transformation

pkg/js/transformer/modules/modules-to-model.ts
```diff
@@ -0,0 +1,191 @@
...
+} from "../../errors";
+import { getTypeLineNumber, getConditionLineNumber, getRelationLineNumber } from "../../util/line-numbers";
+import { constructTransformationError } from "../../util/exceptions";
+
+export interface ModuleFile {
+  name: string;
+  contents: string;
+}
+
+export const transformModuleFilesToModel = (files: ModuleFile[]): Omit<AuthorizationModel, "id"> => {
```
rhamzeh, line 19: Question, should this function also require the fga.mod json? FOr the future so we're not assuming 1.2

pkg/js/transformer/modules/modules-to-model.ts
```diff
@@ -0,0 +1,191 @@
...
+      }
+    }
+  }
+
+  for (const [filename, extended] of Object.entries(extendedTypeDefs)) {
+    const lines = moduleFiles.get(filename)?.split("\n");
+    for (const typeDef of extended) {
+
+      if (!typeDef.relations) {
+        // TODO: Maybe should be an error case or at least a warning?
```
rhamzeh, line 104: That's a good question - I think in the future we may want to allow it, but for not, it makes sense not to allow type extensions that do nothing

pkg/js/transformer/modules/modules-to-model.ts
```diff
@@ -0,0 +1,191 @@
...
+    } catch (error) {
+      if (error instanceof DSLSyntaxError) {
+        errors.push(...error.errors);
+      } else if (error instanceof Error) {
+        errors.push(error as BaseError);
+      }
+    }
+  }
+
+  for (const [filename, extended] of Object.entries(extendedTypeDefs)) {
```
rhamzeh, line 99: Q: do we have validation so that a type cannot be extended twice in the same file?

rhamzeh (approved):

pkg/js/transformer/modules/modules-to-model.ts
ewanharris: Not currently but we should do

pkg/js/transformer/modules/modules-to-model.ts
ewanharris: Yeah I was meaning to add that and forgot.

 I wasn't sure whether to take the whole FGAMod interface or have a second optional parameter that makes it easier to add extras onto in future (like if we ever wanted to allow specifying `ValidationOptions` for the `validateJSON` call)

pkg/js/validator/validate-dsl.ts
```diff
@@ -804,6 +804,7 @@ export function validateJSON(
 
   switch (schemaVersion) {
     case "1.1":
```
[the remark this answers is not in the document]
ewanharris: I guess? Where would the right place for those enums be, in language or in `openfga/api` so they're then part of the SDK/proto?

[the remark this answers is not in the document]
rhamzeh: I would prefer api, but we'd need to discuss that with the core team

pkg/js/transformer/modules/modules-to-model.ts
rhamzeh: 👍🏽 we should have it, but we can create a ticket and defer for after the CLI integration

ewanharris: Split off into #187

pkg/js/transformer/modules/modules-to-model.ts
ewanharris: #188 188

Issue #186, opened 2024-03-08
Map errors raised by `validateJSON` back to their original source files


ewanharris:
When we run the fully combined model through `validateJSON` currently the errors have no metadata associated with them that would allow us to (or at least easily) look up the line and column information associated with the error.

When we raise a validation error we should include the new file and module metadata fields so that we can improve the error reporting for these errors. Then in the catch for `validateJSON` we can lookup the file contents and get line/column data to improve these errors.

ewanharris:
Current status of error types (as listed in [`errors.ts`](https://github.com/openfga/language/blob/feat/148-support-modular-models/pkg/js/errors.ts#L3-L24)):

- "schema-version-required" ❌ wont do, check should be in transform
- "schema-version-unsupported" ❌ wont do, check should be in transform
- "reserved-type-keywords" ✅
- "reserved-relation-keywords" ✅
- "self-error" ❓ this doesn't seem to be referenced anywhere
- "invalid-name" ✅
- "missing-definition" ⚠️ In progress
- "invalid-relation-type" ⚠️ In progress
- "invalid-type" ⚠️ In progress
- "relation-no-entry-point" ❓ Can't do without a third error project
- "tupleuset-not-direct" ⚠️ In progress
- "duplicated-error" ❌ wont do (already handled in transform)
- "allowed-type-schema-10" ❌ wont do
- "assignable-relation-must-have-type" ❓ this doesn't seem to be referenced anywhere
- "allowed-type-not-valid-on-schema-1_0" ❌ wont do
- "invalid-schema" ❌ wont do, check should be in transform
- "invalid-syntax" ❓ this doesn't seem to be referenced anywhere
- "type-wildcard-relation" ⚠️ In progress
- "condition-not-defined" ⚠️ In progress
- "condition-not-used" ⚠️ In progress

PR #191, opened 2024-03-11
feat(js): improve some reported errors from validating final model in modular transform

ewanharris:
## Description

This is a start on mapping the errors raised during a modular transform back to their original source files, it's only a limited set right now but starts the basis of how we could do this.

Given we don't have validation in the Go package yet I'm not really sure what we can do to support those there :/

## References

Start of #186 

## Review Checklist

### DIFF
--- a/pkg/js/tests/modules/module-to-model.test.ts
+++ b/pkg/js/tests/modules/module-to-model.test.ts
@@ -1,4 +1,4 @@
-import { BaseError, ModuleTransformationError } from "../../errors";
+import { ModelValidationSingleError, ModuleTransformationError, ModuleTransformationSingleError } from "../../errors";
 import { transformModuleFilesToModel } from "../../transformer/modules/modules-to-model";
 import { loadModuleTestCases } from "../_testcases";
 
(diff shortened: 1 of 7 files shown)


pkg/js/tests/modules/module-to-model.test.ts
```diff
@@ -21,9 +21,14 @@ describe("transformModuleFilesToModel", () => {
...
-                            let msg = "transformation-error error";
-                            if (err?.line) {
+                        .map((err: ModelValidationSingleError|ModuleTransformationSingleError) => {
+                            let errorType = "transformation-error";
+                            if ((err as ModelValidationSingleError).metadata?.errorType) {
+                                errorType = (err as ModelValidationSingleError).metadata!.errorType;
+                            }
+
+                            let msg = `${errorType} error`;
+                            if (err?.line && !err.metadata) {
```
rhamzeh, line 31: If line and no metadata? Is this expected?

ewanharris: This is working around the fact that the message for some of the errors will have line/column numbers (errors added during transformation) whereas errors added during validation won't.

It's sniffing from the expected errors as we won't be using the `metadata` field for the errors that have line/column numbers in the message as they will all be `transformation-error`, we can't use `instanceof` checks as these aren't ever actually converted into real errors they just have all the properties of the error in the JSON

rhamzeh (approved):

PR #193, opened 2024-03-12
feat: allow passing a custom schema version to modular transform

### LINKED ISSUE #188: Allow providing schema version in transformModuleFilesToModel
Yeah I was meaning to add that and forgot.

 I wasn't sure whether to take the whole FGAMod interface or have a second optional parameter that makes it easier to add extras onto in future (like if we ever wanted to allow specifying `ValidationOptions` for the `validateJSON` call)

_Originally posted by @ewanharris in https://github.com/openfga/language/pull/178#discussion_r1517900934_

ewanharris:
## Description

Adds support for passing the schema version when transforming a modular model.

Initially I was going to do this in an optional manner, but then realised that realistically anyone who is using this will always pass the schema version anyway so it made sense to move this to required and not have a default version.

Validation is left to the standard validation logic.

## References

Closes #188

## Review Checklist

### DIFF
--- a/pkg/go/transformer/module-to-model.go
+++ b/pkg/go/transformer/module-to-model.go
@@ -57,9 +57,12 @@ func (e *ModuleValidationMultipleError) Error() string {
 }
 
 // TransformModuleFilesToModel transforms the provided modules into a singular authorization model.
-func TransformModuleFilesToModel(modules []ModuleFile) (*pb.AuthorizationModel, error) { //nolint:funlen,gocognit,cyclop
+func TransformModuleFilesToModel( //nolint:funlen,gocognit,cyclop
+	modules []ModuleFile,
+	schemaVersion string,
+) (*pb.AuthorizationModel, error) {
 	model := &pb.AuthorizationModel{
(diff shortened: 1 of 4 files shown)


rhamzeh (commented):
Q: why is schema version passed in nested in an options object in JS and directly in Go?

Can we make Go like JS?

ewanharris:
I wasn't 100% decided on what a right signature would be for Go as I think the catchall options object pattern isn't a common pattern like it is in JS.

I initially had it implemented using the [Options pattern](https://dev.to/kittipat1413/understanding-the-options-pattern-in-go-390c) but with making it a required parameter that didn't really work so I went with the parameter. Then most likely in future if we need to add more parameters we could add them using the options pattern with defaults.

I can switch to accepting a struct that copies the JS options if we'd just prefer consistency across the packages?

rhamzeh:
Hmm.. In that case we can make js like Go?

I'm also jot sure maybe schemaVersion should be the first param?

rhamzeh (approved):

PR #198, opened 2024-03-13
chore: correctly handle syntax errors during parsing of the dsl

ewanharris:
## Description

Adds a test for a DSL file with a syntax error and then ensure that we correctly return this as part of the overall modular model transform

## Review Checklist

### DIFF
--- a/pkg/go/transformer/module-to-model.go
+++ b/pkg/go/transformer/module-to-model.go
@@ -81,7 +81,7 @@ func TransformModuleFilesToModel( //nolint:funlen,gocognit,cyclop
 
 		mdl, typeDefExtensions, err := TransformModularDSLToProto(module.Contents)
 		if err != nil {
-			var syntaxError *OpenFgaDslSyntaxMultipleError
+			var syntaxError *multierror.Error
 			if errors.As(err, &syntaxError) {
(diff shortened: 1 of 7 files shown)


pkg/go/transformer/module-to-model.go
```diff
@@ -81,7 +81,7 @@ func TransformModuleFilesToModel( //nolint:funlen,gocognit,cyclop
 
 		mdl, typeDefExtensions, err := TransformModularDSLToProto(module.Contents)
 		if err != nil {
-			var syntaxError *OpenFgaDslSyntaxMultipleError
+			var syntaxError *multierror.Error
```
rhamzeh, line 84: Looks good to me, however pinging @d-jeffery as this change may break some stuff in VS Code handling

rhamzeh (approved):

pkg/go/transformer/module-to-model.go
ewanharris: I think on the JS side there shouldn't be any change as we were correctly handling them already, however because I had the type check wrong in Go syntax errors were being ignored so a transform was appearing successful and just dropping files (you don't want to know how long I wasted trying to figure out why some relations were missing in my CLI testing 😭 )

PR #223, opened 2024-04-02
Make line and column numbers zero based

ewanharris:
## Description

Makes the line and column numbers for the various errors 0 based, we were mostly 1 based already with the exception of some oversights (Go default error in some modular cases) and the line (but not column) on syntax errors from antlr.

This is separated into different commits for the tests and each language to help with review. There is a little bit of awkwardness in that we don't assert the errors in Go and Java due to differences in the errors across languages, but it shouldn't be an issue.

## Review Checklist

### DIFF
--- a/pkg/go/transformer/dsltojson.go
+++ b/pkg/go/transformer/dsltojson.go
@@ -463,7 +463,8 @@ func newOpenFgaDslErrorListener() *OpenFgaDslErrorListener {
 func (c *OpenFgaDslErrorListener) SyntaxError(
 	_ antlr.Recognizer,
 	offendingSymbol interface{},
-	line, column int,
+	line, // line is one based, i.e. the first line will be 1
+	column int, // column is one based, i.e. the first column will be 0
 	msg string,
 	_ antlr.RecognitionException,
 ) {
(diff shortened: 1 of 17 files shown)


rhamzeh (commented):
❤️ - sorry to ask but can you add comments for future us? 🙏🏽 

Re: antlr being (1,0) based and lovely things like that :)

pkg/go/transformer/dsltojson.go
```diff
@@ -463,7 +463,8 @@ func newOpenFgaDslErrorListener() *OpenFgaDslErrorListener {
 func (c *OpenFgaDslErrorListener) SyntaxError(
 	_ antlr.Recognizer,
 	offendingSymbol interface{},
-	line, column int,
+	line, // line is one based, i.e. the first line will be 1
+	column int, // column is one based, i.e. the first column will be 0
```
rhamzeh, line 467:
```suggestion
	column int, // column is zero based, i.e. the first column will be 0
```

pkg/java/src/main/java/dev/openfga/language/OpenFgaDslErrorListener.java
```diff
@@ -23,6 +23,8 @@ public List<SyntaxError> getErrors() {
     }
 
     @Override
+    // line is one based, i.e. the first line will be 1
+    // column is one based, i.e. the first column will 0
```
rhamzeh, line 27:
```suggestion
    // column is zero based, i.e. the first column will 0
```

rhamzeh (approved):

PR #228, opened 2024-04-09
Prevent Directory Traversal in YAML Content Validation

### LINKED ISSUE #199: Ensure that every file referenced in the `fga.mod` is in the current or sub-directory
In [fga.mod transform](https://github.com/openfga/language/blob/feat/148-support-modular-models/pkg/js/transformer/modules/mod-to-json.ts#L165-L184) validate that a path contains no `../` that would lead to a file being located out of the current directory

Siddhant-K-code:
## Description

This PR fixes a vulnerability in the YAML content validation process. The changes ensure that paths containing `../` are not allowed, preventing directory traversal attacks.

## References

fixes #199

## Review Checklist

### DIFF
--- a/tests/data/fga-mod-transformer-cases.yaml
+++ b/tests/data/fga-mod-transformer-cases.yaml
@@ -172,3 +172,58 @@
       column:
         start: 4
         end: 12
+- name: contents path contains dot-dot-slash sequence and other invalid paths
+  modFile: |
+    schema: "1.2"
+    contents:
+      - core.fga
+      - ../team1/project.fga
+      - team2/../team3/board.fga
+      - team2\\..\\team3\\board.fga
+      - /absolute/path.fga
+      - "%2e%2e%2fteam4%2fproject.fga"
+      - "%2e%2e%5cteam4%5cproject.fga"
+      - wiki.fga
+  expected_errors:
+    - msg: invalid contents item ../team1/project.fga
+      line:
+        start: 3
+        end: 3
+      column:
+        start: 4
+        end: 24
+    - msg: invalid contents item team2/../team3/board.fga
+      line:
+        start: 4
+        end: 4
+      column:
+        start: 4
+        end: 28
+    - msg: invalid contents item team2\\..\\team3\\board.fga
+      line:
+        start: 5
+        end: 5
+      column:
+        start: 4
+        end: 31
+    - msg: invalid contents item /absolute/path.fga
+      line:
+        start: 6
+        end: 6
+      column:
+        start: 4
+        end: 22
+    - msg: invalid contents item %2e%2e%2fteam4%2fproject.fga
+      line:
+        start: 7
+        end: 7
+      column:
+        start: 4
+        end: 34
+    - msg: invalid contents item %2e%2e%5cteam4%5cproject.fga
+      line:
+        start: 8
+        end: 8
+      column:
+        start: 4
+        end: 34
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -32,4 +32,6 @@ updates:
      dependencies:
         patterns:
           - "*"
+        exclude-patterns:
+          - "eslint"
 
--- a/.github/workflows/main-go.yaml
+++ b/.github/workflows/main-go.yaml
@@ -49,7 +49,7 @@ jobs:
         fetch-depth: 0
 
     - name: Set up Docker Buildx
-      uses: docker/setup-buildx-action@2b51285047da1547ffb1b2203d8be4c0af6b1f20 # v3.2.0
+      uses: docker/setup-buildx-action@d70bba72b1f3fd22344832f00baa16ece964efeb # v3.3.0
 
     - uses: actions/setup-go@0c52d547c9bc32b1aa3301fd7a9cb496313a4491 # v5.0.0
       with:
--- a/.github/workflows/main-java.yml
+++ b/.github/workflows/main-java.yml
@@ -35,7 +35,7 @@ jobs:
             fetch-depth: 0
 
         - name: Set up Docker Buildx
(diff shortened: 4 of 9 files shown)


ewanharris:
Thanks for the PR @Siddhant-K-code, I've just pushed #229 which should fix the CI failure we're seeing.

If you're also comfortable with handling the Go side of this fix, would you mind adding a test? It would go inside [this file](https://github.com/openfga/language/blob/main/tests/data/fga-mod-transformer-cases.yaml) (as our tests are generated from the yaml files it will be automatically added to the Go suite so would need a fix there too)

Siddhant-K-code:
@ewanharris I've committed d086abf to add the tests, is that the only thing required, do I need to run any script to generate test files?

Also, will rebase my PR after #229 is merged.

ewanharris:
Thanks for adding that! As we've added the test it's now being ran against the Go language package, would you mind also applying the fix over there?

I'm also wondering now whether a check for `../` is enough:

* We should also consider Windows path separators (i.e. `..\\`) as I don't believe we have a requirement to use posix separators in our [API validation](https://github.com/openfga/api/blob/main/openfga/v1/authzmodel.proto#L96)
* I think if we're truly wanting to round this out, we also need to handle replacing URI encoded characters to prevent someone writing `%2e%2e%2fname.fga` (I believe Node's `fs` doesn't handle this but I'm not sure if the APIs we use in VS Code and potentially other environments do).

Siddhant-K-code:
> Thanks for adding that! As we've added the test it's now being ran against the Go language package, would you mind also applying the fix over there?
> 

Will need to explore that. Can you point to any code reference, anyway i would also checkout it.

> I'm also wondering now whether a check for `../` is enough:
> 
> * We should also consider Windows path separators (i.e. `..\\`) as I don't believe we have a requirement to use posix separators in our [API validation](https://github.com/openfga/api/blob/main/openfga/v1/authzmodel.proto#L96)
> * I think if we're truly wanting to round this out, we also need to handle replacing URI encoded characters to prevent someone writing `%2e%2e%2fname.fga` (I believe Node's `fs` doesn't handle this but I'm not sure if the APIs we use in VS Code and potentially other environments do).

Yeah, there could be many cases like this 🫠. It would be tricky to handle all.

ewanharris:
The Go package lives in this same repo under [pkg/go](https://github.com/openfga/language/tree/main/pkg/go), the code block we'd need to add the check to is [here](https://github.com/openfga/language/blob/main/pkg/go/transformer/mod-to-json.go#L125-L137).

For the Windows/encoded characters piece, I think if we did something like below that should tick off the different variants, (I also added an absolute path check as that was also been raised as part of this).

```js
// replace URI encoded characters
let filename = file.value.replace(/%2e/ig, ".");
filename = filename.replace(/%2f/ig, "/");
filename = filename.replace(/%5c/ig, "\\");

// error if the filename indicates a potential directory traversal or absolute path
if (filename.includes("../") || filename.includes("..\\" ) || filename.startsWith("/")) {
    // error
}
```

Siddhant-K-code:
Tried some things in 00aa1a4

Siddhant-K-code:
Ahh, tests are failing now ;_;

Will debug it later 🫠

tests/data/fga-mod-transformer-cases.yaml
```diff
@@ -172,3 +172,58 @@
...
+  modFile: |
+    schema: "1.2"
+    contents:
+      - core.fga
+      - ../team1/project.fga
+      - team2/../team3/board.fga
+      - team2\\..\\team3\\board.fga
+      - /absolute/path.fga
+      - %2e%2e%2fteam4%2fproject.fga
+      - %2e%2e%5cteam4%5cproject.fga
```
ewanharris, line 185: It looks like we're getting some yaml parse errors due to these starting with `%`, so we probably have to wrap them in quotes

```
* validation error at line=8, column=5: Plain value cannot start with directive indicator character % at line 8, column 5:
- %2e%2e%2fteam4%2fproject.fga
    ^
* validation error at line=9, column=5: Plain value cannot start with directive indicator character % at line 9, column 5:
- %2e%2e%5cteam4%5cproject.fga
    ^
```

```suggestion
      - "%2e%2e%2fteam4%2fproject.fga"
      - "%2e%2e%5cteam4%5cproject.fga"
```

Siddhant-K-code: Accepted the change, now will require an approval to run workflows

Siddhant-K-code:
There seem to be more problems in this. 🤔 
Will check it after my dinner

Siddhant-K-code:
@ewanharris I am not sure, how are other tests are getting affected 🤔 Unrelated changes functions' tests also failing, can you help me out?

image

ewanharris (commented):
It looks like the test failing in Go was due to changing from the if/else if block.

Suggested some edits but feel free to apply locally and commit as I'm wary that the suggestions might cause some linting errors around spacing 😅

Siddhant-K-code:
Thanks a lot, @ewanharris for helping me out, I've tested the tests, and builds in my Gitpod workspace, and it seems to be passing now. Requires final approval to run the GHA workflows.

ewanharris (approved):
Thanks for tackling this @Siddhant-K-code!

Siddhant-K-code:
ahhh yehahhhh,  loved the green CI :)

ewanharris:
I just pushed to the branch in order to squash the commit, thanks again for working through this @Siddhant-K-code, it's great to see your contributions!

Siddhant-K-code:
Thank you so much, @ewanharris ! I really appreciate your help and your kind words. It's been a pleasure contributing to all the OpenFGA projects, and I'm glad to have your support. Looking forward to more future collaborations ❤️

PR #233, opened 2024-04-11
fix: syntax error for missing module name

d-jeffery:
## Description

* Easiest way to deal with this is with a generic syntax error
* Added its own test data package due to syntax errors causing a parsing halt

## Review Checklist

### DIFF
--- a/pkg/js/transformer/dsltojson.ts
+++ b/pkg/js/transformer/dsltojson.ts
@@ -108,6 +108,10 @@ class OpenFgaDslListener extends OpenFGAListener {
   private rewriteStack: StackRelation[] = [];
 
   exitModuleHeader = (ctx: ModuleHeaderContext) => {
+    if (!ctx._moduleName) {
+      return;
+    }
+
     this.isModularModel = true;
     this.moduleName = ctx._moduleName.getText();
   };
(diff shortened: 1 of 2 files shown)


rhamzeh (approved):
Can you add another test for 5 whitespaces after module and no name? It will fail the same way but just to have a test for that

```
moduleWSWSWSWSWS

type user
```

tests/data/transformer-module/06-module-with-syntax-error/module/incomplete.fga
```diff
@@ -0,0 +1,3 @@
+module
```
ewanharris, line 1: Could we maybe put this under `dsl-syntax-validation-cases.yaml` rather than the module transform tests? (I'm guessing that would then require the fix to be implemented across Go and Java too)

d-jeffery: Currently Java & Go listeners dont catch on a missing module names;

https://github.com/openfga/language/blob/e4a4d6ab4e1c87a8e6f93301605f8b485ee4baac/pkg/java/src/main/java/dev/openfga/language/OpenFgaDslListener.java#L67-L69

https://github.com/openfga/language/blob/f45e80a9f169370f7cbeb5da9d147e5ff0e7305e/pkg/go/transformer/dsltojson.go#L103-L109

ewanharris: Nice!

ewanharris (approved):

rhamzeh (approved):

Issue #278, opened 2024-06-18
Add support for validating the store file


rhamzeh:
The Store file validation is now currently done by both VS Code and the CLI. We should move it here and export it so that VS Code, the CLI and IntelliJ can make use of it

Issue #226, opened 2024-04-04
add methods to validate types, relations, ids and tuples


rhamzeh:
> Let's plan on moving this logic to language and have test cases for them, we can then implement this in several languages consistently

_Originally posted by @rhamzeh in https://github.com/openfga/vscode-ext/pull/215#pullrequestreview-1981267968_
            
Current set of regexes used:

Protos:
- Type: https://github.com/openfga/api/blob/main/openfga/v1/openfga.proto#L19 | `"^[^:#@\\s]{1,254}$"`
- Relation: https://github.com/openfga/api/blob/main/openfga/v1/openfga.proto#L68 | `"^[^:#@\\s]{1,50}$"`
- Condition: https://github.com/openfga/api/blob/main/openfga/v1/openfga.proto#L39C16-L39C33 | `"^[^\\s]{2,256}$"` 
- Object: https://github.com/openfga/api/blob/main/openfga/v1/openfga.proto#L80 | `"^[^\\s]{2,256}$"`

API:
https://github.com/openfga/openfga/blob/main/pkg/tuple/tuple.go#L35-L38
```
	userIDRegex   = regexp.MustCompile(`^[^:#\s]+$`)
	objectRegex   = regexp.MustCompile(`^[^:#\s]+:[^#:\s]+$`)
	userSetRegex  = regexp.MustCompile(`^[^:#\s]+:[^#\s]+#[^:#\s]+$`)
	relationRegex = regexp.MustCompile(`^[^:#@\s]+$`)
```

Validation has to pass both sets of regexes above in order to be valid according to the API.

---

On the VS Code/Playground side we use a different set of less permissive regexes:
https://github.com/openfga/vscode-ext/blob/07e65ccd3a0d533bcbbb2b4495899d73d2656e3b/server/src/openfga-yaml-schema.ts#L126-L154

We have yet more regexes in language:
https://github.com/openfga/language/blob/35310a2e97386126a01db1c4360e013d984f98f2/pkg/js/validator/validate-dsl.ts#L7-L10

PR #286, opened 2024-07-01
chore: moved store validation to language

d-jeffery:
## References

- closes https://github.com/openfga/language/issues/278 

## Review Checklist

### DIFF
--- a/pkg/js/tests/validate-rules.test.ts
+++ b/pkg/js/tests/validate-rules.test.ts
@@ -1,0 +1,167 @@
+import { describe, expect, it } from "@jest/globals";
+import { Validate } from "../validator/validate-rules";
+
+// These tests are a subset of bad formats that all validation rules that should fail
+const validatedBadStructure = (validator: any) => {
+  it("should fail if '::' is the delimiter", () => {
+    expect(validator("item::1")).toBeFalsy();
+  });
+
+  it("should fail if ':' at start", () => {
+    expect(validator(":item:1")).toBeFalsy();
+  });
+
+  it("should fail if ':' at end", () => {
+    expect(validator("item:1:")).toBeFalsy();
+  });
+
+  it("should fail if '#' is present, but not ':'", () => {
+    expect(validator("item#relation")).toBeFalsy();
+  });
+
+  it("should fail if '##' is a delimiter", () => {
+    expect(validator("item:1##relation")).toBeFalsy();
+  });
+
+  it("should fail if '#' at start", () => {
+    expect(validator("#item:1")).toBeFalsy();
+  });
+
+  it("should fail if '#' at end", () => {
+    expect(validator("item:1#")).toBeFalsy();
+  });
+
+  it("should fail if '#' is before ':'", () => {
+    expect(validator("ite#m:1")).toBeFalsy();
+  });
+
+  it("should fail if wildcard present in object name", () => {
+    expect(validator("it*em:1")).toBeFalsy();
+  });
+
+  it("should fail if wildcard at start", () => {
+    expect(validator("*:1")).toBeFalsy();
+  });
+
+  it("should fail if wildcard has no ':'", () => {
+    expect(validator("item*thing")).toBeFalsy();
+  });
+
+  it("should fail if wildcard is proceeded by text", () => {
+    expect(validator("item:*thing")).toBeFalsy();
+  });
+
+  it("should fail if wildcard appears multiple times", () => {
+    expect(validator("item:**")).toBeFalsy();
+  });
+};
+
+describe("Validation Rules", () => {
+  describe("Rule 'Object'", () => {
+    it("should pass '<object>:<id>'", () => {
+      expect(Validate.object("document:1")).toBeTruthy();
+    });
+
+    it("should fail if no ':' delimiter", () => {
+      expect(Validate.object("document1")).toBeFalsy();
+    });
+
+    it("should fail if includes relation", () => {
+      expect(Validate.object("document:1#relation")).toBeFalsy();
+    });
+
+    validatedBadStructure(Validate.object);
+  });
+
+  describe("Rule 'User'", () => {
+    it("shoud pass for 'UserSet'", () => {
+      expect(Validate.user("group:engineering#member")).toBeTruthy();
+    });
+
+    it("shoud pass for 'UserObject'", () => {
+      expect(Validate.user("group:engineering")).toBeTruthy();
+    });
+
+    it("shoud pass for 'UserWildcard'", () => {
+      expect(Validate.user("group:*")).toBeTruthy();
+    });
+
+    it("shoud fail when missing <id>'", () => {
+      expect(Validate.user("group")).toBeFalsy();
+    });
+
+    validatedBadStructure(Validate.user);
+  });
+
+  describe("Rule 'UserSet'", () => {
+    it("should pass '<type>:<id>#<relation>'", () => {
+      expect(Validate.userSet("group:engineering#member")).toBeTruthy();
+    });
+
+    it("shoud fail for 'UserObject'", () => {
+      expect(Validate.userSet("group:engineering")).toBeFalsy();
+    });
+
+    it("shoud fail for 'UserWildcard'", () => {
+      expect(Validate.userSet("group:*")).toBeFalsy();
+    });
+
+    it("should fail if missing <id>", () => {
+      expect(Validate.userSet("group")).toBeFalsy();
+    });
+
+    validatedBadStructure(Validate.userSet);
+  });
+
+  describe("Rule 'UserObject'", () => {
+    it("should pass '<type>:<id>", () => {
+      expect(Validate.userObject("group:engineering")).toBeTruthy();
+    });
+
+    it("should fail if contains '#", () => {
+      expect(Validate.userObject("group:engineering#member")).toBeFalsy();
+    });
+
+    it("should fail if 'Wildcard' is present", () => {
+      expect(Validate.userObject("group:*")).toBeFalsy();
+    });
+
+    it("should fail if missing <id>", () => {
+      expect(Validate.userObject("group")).toBeFalsy();
+    });
+
+    validatedBadStructure(Validate.userObject);
+  });
+
+  describe("Rule 'UserWildcard'", () => {
+    it("should pass '<type>:*", () => {
+      expect(Validate.userWildcard("group:*")).toBeTruthy();
+    });
+
+    it("should fail if missing 'UserObject'", () => {
+      expect(Validate.userWildcard("group:organization")).toBeFalsy();
+    });
+
+    it("should fail if missing '*' delimiter", () => {
+      expect(Validate.userWildcard("group")).toBeFalsy();
+    });
+
+    validatedBadStructure(Validate.userWildcard);
+  });
+
+  describe("Rule 'Type'", () => {
+    it("should pass '<type>", () => {
+      expect(Validate.type("folder")).toBeTruthy();
+    });
+
+    it("should fail 'UserObject", () => {
+      expect(Validate.type("folder:1")).toBeFalsy();
+    });
+
+    it("should fail 'UserSet'", () => {
+      expect(Validate.type("folder:1#relation")).toBeFalsy();
+    });
+
+    validatedBadStructure(Validate.type);
+  });
+});
--- a/pkg/js/validator/validate-rules.ts
+++ b/pkg/js/validator/validate-rules.ts
@@ -1,0 +1,49 @@
+export const Rules = {
+  type: "[^:#@\\*\\s]{1,254}",
+  relation: "[^:#@\\*\\s]{1,50}",
+  condition: "[^\\*\\s]{2,256}",
+  id: "[^#:\\*\\s]+",
+  object: "[^\\s]{2,256}",
+};
+
+export const Validate = {
+  // An Object is composed of a type and identifier (e.g. 'document:1')
+  object: (object: string): boolean => {
+    return validateFieldValue(`^${Rules.type}:${Rules.id}$`, object) && validateFieldValue(`^${Rules.object}$`, object);
+  },
+  // Relation reference
+  relation: (relation: string): boolean => {
+    return validateFieldValue(`^${Rules.relation}$`, relation);
+  },
+  // User is composed of type, identifier and relation (e.g. 'group:engineering#member')
+  userSet: (user: string): boolean => {
+    return validateFieldValue(`^${Rules.type}:${Rules.id}#${Rules.relation}$`, user);
+  },
+  // User is composed of type and identifier (e.g. 'group:engineering')
+  userObject: (userObject: string): boolean => {
+    return (
+      validateFieldValue(`^${Rules.type}:${Rules.id}$`, userObject) &&
+      validateFieldValue(`^${Rules.object}$`, userObject)
+    );
+  },
+  // User is composed of type, and a wildcard (e.g. 'group:*')
+  userWildcard: (userWildcard: string): boolean => {
+    return validateFieldValue(`^${Rules.type}:\\*$`, userWildcard);
+  },
+  // Is either a userset, userobject or a user wildcard
+  user: (user: string): boolean => {
+    return Validate.userSet(user) || Validate.userObject(user) || Validate.userWildcard(user);
+  },
+  // Condition name
+  relationshipCondition: (condition: string) => {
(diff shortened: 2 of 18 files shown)


pkg/js/tests/validate-rules.test.ts
```diff
@@ -0,0 +1,70 @@
+import { describe, expect, it } from "@jest/globals";
+import { Validate } from "../validator/validate-rules";
+
+describe("validation rules", () => {
+  describe("Rule 'Object'", () => {
+    it("should pass '<object>:<id>'", () => {
+      expect(Validate.object("document:1")).toBeTruthy();
+    });
+
+    it("should fail if no ':' delimiter", () => {
```
rhamzeh, line 10: Can you add more test cases?

These should fail
* document::1
* :document:1
* document:1:
* document#1
* document:1#relation

pkg/js/tests/validate-rules.test.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  describe("Rule 'User'", () => {
+    it("shoud pass for 'UserSet'", () => {
+      expect(Validate.user("group:engineering#member")).toBeTruthy();
+    });
+    it("shoud pass for 'UserObject'", () => {
+      expect(Validate.user("group:engineering")).toBeTruthy();
+    });
+    it("shoud pass for 'UserWildcard'", () => {
+      expect(Validate.user("group:*")).toBeTruthy();
```
rhamzeh, line 23: Same here, can we add more test cases?

pkg/js/tests/validate-rules.test.ts
```diff
@@ -0,0 +1,70 @@
...
+      expect(Validate.userObject("group#member")).toBeFalsy();
+    });
+  });
+
+  describe("Rule 'UserWildcard'", () => {
+    it("should pass '<type>:*", () => {
+      expect(Validate.userWildcard("group:*")).toBeTruthy();
+    });
+
+    it("should fail if missing '*' delimiter", () => {
```
rhamzeh, line 60: should fail e.g. `user:anne`

And add more test cases like above

pkg/js/tests/validate-rules.test.ts
```diff
@@ -0,0 +1,70 @@
...
+      expect(Validate.userWildcard("group")).toBeFalsy();
+    });
+  });
+
+  describe("Rule 'Type'", () => {
+    it("should pass '<type>", () => {
+      expect(Validate.type("folder")).toBeTruthy();
```
rhamzeh, line 67: Can you add the ones that would fail?

pkg/js/validator/validate-rules.ts
```diff
@@ -0,0 +1,46 @@
+export const Rules = {
+  type: "[^:#@\\s]{1,254}",
+  relation: "[^:#@\\s]{1,50}",
+  condition: "[^\\s]{2,256}",
+  object: "[^\\s]{2,256}",
+  id: "[^#:\\s]+",
+};
```
rhamzeh, line 7: If we have any validation on the server for the condition param name, let's add it here

pkg/js/validator/validate-rules.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export const Validate = {
+  // An Object is composed of a type and identifier (e.g. 'document:1')
+  object: (object: string): boolean => {
+    return validateFieldValue(`^${Rules.object}:${Rules.id}$`, object);
```
rhamzeh, line 12:
```suggestion
    return validateFieldValue(`^${Rules.type}:${Rules.id}$`, object);
```

This should ALSO pass Rules.object, but Rules.object is for the full thing

pkg/js/validator/validate-rules.ts
```diff
@@ -0,0 +1,46 @@
...
+  relation: (relation: string): boolean => {
+    return validateFieldValue(`^${Rules.relation}$`, relation);
+  },
+  // User is composed of type, identifier and relation (e.g. 'group:engineering#member')
+  userSet: (user: string): boolean => {
+    return validateFieldValue(`^${Rules.type}:${Rules.id}#${Rules.relation}$`, user);
+  },
+  // User is composed of type and identifier (e.g. 'group:engineering')
+  userObject: (userObject: string): boolean => {
+    return validateFieldValue(`^${Rules.type}:${Rules.id}$`, userObject);
```
rhamzeh, line 24: This should redirect to validate against object

pkg/js/validator/validate-store.ts
```diff
@@ -0,0 +1,976 @@
...
+// Validate Relation
+function validateRelation(tuple: TupleKey, typeDefs: TypeDefinition[], instancePath: string): boolean {
+  const errors = [];
+
+  // Check if relation exists on given type
+  let doesExistOnType = false;
+  if (tuple.user.includes("#")) {
+    const user = tuple.user.split(":")[0];
+    const userRelation = tuple.user.split("#")[1];
+    for (const typeDef of typeDefs) {
```
rhamzeh, line 203: In a follow-up PR, let's pre-compute these. The language representation of the model should have all the relations per type precomputed so as to improve performance.

pkg/js/validator/validate-store.ts
```diff
@@ -0,0 +1,976 @@
...
+function validateTypeRestrictions(
+  tuple: TupleKey,
+  typeDefs: TypeDefinition[],
+  conditions: Record<string, Condition> | undefined,
+  instancePath: string,
+): boolean {
+  validateTuple.errors = validateTuple.errors || [];
+
+  const mappedTuple = mapTuple(tuple);
+  const object = tuple.object.split(":")[0];
```
rhamzeh, line 259: What is this doing? It seems to be getting the type, not the object

pkg/js/validator/validate-store.ts
```diff
@@ -0,0 +1,976 @@
...
+        return false;
+      }
+    }
+  }
+  return true;
+}
+
+// Validation for tuples
+const validateTuple: SchemaValidateFunction = function (
+  this: { jsonModel: AuthorizationModel },
```
rhamzeh, line 324: Any particular reason you are using `this` as the variable name?

pkg/js/validator/validate-store.ts
```diff
@@ -0,0 +1,976 @@
...
+  tuple: TupleKey,
+  cxt: { instancePath: string },
```
rhamzeh, line 326:
```suggestion
  ctx: { instancePath: string },
```

pkg/js/validator/validate-store.ts
```diff
@@ -0,0 +1,976 @@
...
+    validateRelation(tuple, jsonModel.type_definitions, cxt.instancePath) &&
+    validateConditionExists(tuple, jsonModel.conditions, cxt.instancePath) &&
+    validateTypeRestrictions(tuple, jsonModel.type_definitions, jsonModel.conditions, cxt.instancePath)
+  );
+};
+
+// Validate the user field for check and list_objects
+function validateUserField(model: AuthorizationModel, types: string[], userField: string, instancePath: string) {
+  const errors = [];
+  const user = userField.split(":")[0];
```
rhamzeh, line 353: This is also getting the type, not the object

pkg/js/validator/validate-store.ts
```diff
@@ -0,0 +1,976 @@
...
+  const context = listObjects.context;
+  for (const testParam in context) {
+    if (!params.includes(testParam)) {
+      errors.push(unidentifiedTestParam(testParam, instancePath + `/context/${testParam}`));
+    }
+  }
+  return errors;
+}
+
+// Validate List User
```
rhamzeh, line 509: List Users

pkg/js/validator/validate-store.ts
```diff
@@ -0,0 +1,976 @@
...
+  instancePath: string,
+) {
+  const errors = [];
+
+  const types = model.type_definitions.map((d) => d.type);
+
+  if (listUsers && isStringValue(listUsers.object)) {
+    const listUserObj = listUsers.object;
+
+    const object = listUserObj.split(":")[0];
```
rhamzeh, line 524:
```suggestion
    const type = listUserObj.split(":")[0];
```

rhamzeh: Considering this is used everywhere, can we make this a helper?

Have a tuple parser that converts from a string to the required fields, so that this and similar calls can become: `const { type } = parseObjectString(listUserObj)`


And `parseObjectString` would return `{ type, id }`
And `parseUserString` would return `{ type, id, relation, wildcard }`

pkg/js/validator/validate-store.ts
```diff
@@ -0,0 +1,976 @@
...
+    if (!errors.length) {
+      if (!tuples.map((tuple) => tuple.object).filter((object) => object === listUserObj).length) {
+        errors.push(undefinedTypeTuple(listUserObj, instancePath + "/object"));
+      }
+    }
+  }
+
+  // Check user fileter
+  if (listUsers.user_filter) {
+    for (const typeNo in listUsers.user_filter) {
```
rhamzeh, line 540: Can you call it userFilter instead? I'm not sure what typeNo is an abbreviation of

pkg/js/validator/validate-store.ts
```diff
@@ -0,0 +1,976 @@
...
+
+  // Collect params for validity checking
+  const params: string[] = [];
+  for (const condition in model.conditions) {
+    for (const param in model.conditions[condition].parameters) {
+      params.push(param);
+    }
+  }
+
+  for (const testNo in store.tests) {
```
rhamzeh, line 609: call it singleTest instead of testNo

pkg/js/validator/validate-store.ts
```diff
@@ -0,0 +1,976 @@
...
+    if (!test) {
+      continue;
+    }
+
+    if (test.tuples && test.tuples.length) {
+      tuples.push(...test.tuples);
+    }
+
+    // Validate check
+    for (const checkNo in test.check) {
```
rhamzeh, line 626:
```suggestion
    for (const singleCheckTest of test.check) {
```

pkg/js/validator/validate-store.ts
```diff
@@ -0,0 +1,976 @@
...
+
+  for (const testNo in store.tests) {
+    // Collect valid tuples
+    const tuples = [];
+    if (store.tuples && store.tuples.length) {
+      tuples.push(...store.tuples);
+    }
+
+    const test = store.tests[testNo];
+    if (!test) {
```
rhamzeh, line 617: How can this be false? You're looping over the store tests already

pkg/js/validator/validate-store.ts
```diff
@@ -0,0 +1,976 @@
...
+          test.check[checkNo],
+          tuples,
+          params,
+          instancePath + `/tests/${testNo}/check/${checkNo}`,
+        ),
+      );
+    }
+
+    // Validate list objects
+    for (const listNo in test.list_objects) {
```
rhamzeh, line 642:
```suggestion
    for (const singleListObjectsTest of test.list_objects) {
```

rhamzeh: That way you don't need to do `if !test.list_objects[listNo]` below, just
```
if !singleListObjectsTest.user || !singleListObjectsTest.type)
```

pkg/js/validator/validate-store.ts
rhamzeh: With the appropriate changes below

pkg/js/validator/validate-store.ts
```diff
@@ -0,0 +1,976 @@
...
+          model,
+          test.list_objects[listNo],
+          tuples,
+          params,
+          instancePath + `/tests/${testNo}/list_objects/${listNo}`,
+        ),
+      );
+    }
+
+    for (const listNo in test.list_users) {
```
rhamzeh, line 657: Same comment as above, use `for of` + singleListUsersTest

pkg/js/validator/validate-store.ts
```diff
@@ -0,0 +1,976 @@
...
+
+  if (errors.length) {
+    validateStore.errors?.push(...errors);
+    return false;
+  }
+  return true;
+}
+
+const validateStore: SchemaValidateFunction = function (
+  this: { jsonModel: AuthorizationModel },
```
rhamzeh, line 681: this and ctx, can we have better names?

pkg/js/validator/validate-store.ts
```diff
@@ -0,0 +1,976 @@
...
+      type: "object",
+      schema: false,
+      errors: true,
+      validate: validateTuple,
+    })
+    .compile(OPENFGA_TUPLES_SCHEMA);
+}
+
+// YAML tuple validation
+const OPENFGA_TUPLES_SCHEMA: Schema = {
```
rhamzeh, line 727: You might want to have this in a separate file if you want to re-use this across go, js and Java. But you can move it out in a separate PR

pkg/js/tests/validate-rules.test.ts
d-jeffery: Separating test cases into 2 groups, those specific to a validator, and those which are just checks for bad formats that should universally fail.

pkg/js/validator/validate-store.ts
d-jeffery: It's an argument passed in during invocation of `.call()`

https://github.com/openfga/language/pull/286/files#diff-545133f119b2a360e2aa4a2c410f1a5e39fe617ab5377005c29f96f42ce61ae9R32

This was taken from documentation I read when I originally wrote it. I'll try to simplify and remove the `this` assignment.

pkg/js/validator/validate-store.ts
d-jeffery: Renamed variables.

pkg/js/validator/validate-store.ts
d-jeffery: https://github.com/openfga/language/issues/290

pkg/js/validator/validate-store.ts
d-jeffery: I've changed over the names - `cxt` is the parameter name from ajv. 

https://ajv.js.org/keywords.html#schema-compilation-context

pkg/js/validator/validate-store.ts
d-jeffery: Renamed

pkg/js/validator/validate-store.ts
d-jeffery: Removed.

pkg/js/validator/validate-store.ts
d-jeffery: Problem with `of` is that it throws a `is not iterable` exception on **undefined** values, where as `in` doesn't.

Happy to make changes to use `of`, I'll just have to null-check ahead of the loop - is that preferable for you?

pkg/js/validator/validate-store.ts
d-jeffery: Renamed

pkg/js/validator/validate-store.ts
d-jeffery: `jsonModel` is needed for both the validation of the overall store, as well as the tuples.

I'm setting `this` here, via `call`, such that the `jsonModel` is persisted for all validation being referenced by `validateStore` & `validateTuples`. 

Doing so means we dont have to rebuild the schema validator every time we want to invoke  the validation.

I've tried today to do this several alternatives today, and I cant find a way to provide the model data.

pkg/js/validator/validate-store.ts
rhamzeh: Isn't this just using [JSON Schema](https://json-schema.org/)? That's a standard so should be applicable for Go and Java too.

See https://json-schema.org/implementations

E.g.
Java: https://github.com/networknt/json-schema-validator
Go: https://github.com/kaptinlin/jsonschema

pkg/js/validator/validate-rules.ts
d-jeffery: Is there a reason to keep `Rules.object` definition then? It appears to only be used to enforce string length, and that it includes no spaces.

Both `Rules.type` and `Rules.id` are more strict and enforce no spacing. Can I just add length check to the strings if that is the case?

d-jeffery: Also does the `id` have a max length, or is this just checked by `Rules.object`

rhamzeh (approved):
FYI I removed the reference to https://github.com/openfga/language/issues/226 from the description

PR #308, opened 2024-08-08
feat: support direct assignments in graph

miparnisari:
Follow-up of https://github.com/openfga/language/pull/307.

I'm building this incrementally. More PRs will follow.

### DIFF
--- a/pkg/go/graph/graph_builder_test.go
+++ b/pkg/go/graph/graph_builder_test.go
@@ -41,6 +41,106 @@ rankdir=BT
 2 -> 1 [label=direct];
 }`,
 		},
+		`direct_assignment_with_wildcard`: {
+			model: `
+				model
+					schema 1.1
+				type folder
+					relations
+						define viewer: [user:*]
+				type user`,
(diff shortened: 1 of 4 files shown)


pkg/go/graph/graph_builder_test.go
```diff
@@ -37,6 +37,106 @@ rankdir=BT
...
+			expectedOutput: `digraph {
+graph [
+rankdir=BT
+];
+
+// Node definitions.
+0 [label=folder];
+1 [label="folder#viewer"];
+2 [label=user];
+
```
jpadilla, line 139: this file will get gnarly pretty quick

jpadilla (approved):

pkg/go/graph/graph_builder_test.go
miparnisari: Yes. Also it's hard to quickly verify what the DOT looks like. I will move all the tests to a separate folder later on, with .FGA files and .DOT files. But not yet

PR #310, opened 2024-08-08
feat: support TTUs in graph

miparnisari:


pkg/go/graph/graph_builder_test.go
```diff
@@ -190,6 +190,110 @@ rankdir=BT
...
+0 [label=document];
+1 [label="document#parent"];
+2 [label=folder];
+3 [label="document#viewer"];
+4 [label="folder#admin"];
+5 [label=user];
+
+// Edge definitions.
+2 -> 1 [label=direct];
+4 -> 3 [headlabel="(document#parent)"];
```
elbuo8, line 222: Why make it a label, rather than actually connect the edges? `document#parent` is not an independent node when it comes to TTU.

miparnisari: What edges would you connect?

miparnisari: image

elbuo8 (approved):

elbuo8 (approved):

jpadilla (approved):

PR #323, opened 2024-08-16
test: add more tests for graph

miparnisari:
Reviewers: please read every model in this folder: https://github.com/openfga/language/tree/main/tests/data/transformer 

and then use https://dreampuf.github.io/GraphvizOnline/ to visualize the graph to confirm that it's what you expect.


ewanharris (approved):

ewanharris:
It took some thinking to get my head around some of the graphs (I found `92-expenses` quite awkward) but things look good to me. I think it would be good to have more sets of eyes on this though

miparnisari:
@ewanharris, yes, the drawing is very hard to visualize when there are nested usersets involved :(

PR #342, opened 2024-09-12
feat: graph now has edges for computed usersets

miparnisari:
## Description

This change makes it so that

```
define x: y
define y: x
```

are marked in the graph as two _dashed_ edges representing the type "computed". 

But this model

```
define x: y or z
define y: x or z
define z: x or y
```

only has edges of the "rewrite" type.


pkg/go/graph/graph_edge.go
```diff
@@ -35,7 +36,7 @@ func (n *AuthorizationModelEdge) Attributes() []encoding.Attribute {
 		})
 	}
 
-	if n.edgeType == RewriteEdge {
+	if n.edgeType == ComputedEdge {
```
adriantam, line 39: not for this PR but curious on why we don't use switch / case?

adriantam (approved):

pkg/go/graph/graph_edge.go
miparnisari: No good reason 😆  i can send a follow-up

PR #344, opened 2024-09-13
feat: add cycle detection

adriantam:
## Description
Add cycle detection in go graph library.  This serves as future starting point for performance optimization in OpenFGA server.

## Review Checklist

### DIFF
--- a/pkg/go/graph/graph.go
+++ b/pkg/go/graph/graph.go
@@ -3,9 +3,11 @@ package graph
 import (
 	"errors"
 
+	"gonum.org/v1/gonum/graph"
 	"gonum.org/v1/gonum/graph/encoding"
 	"gonum.org/v1/gonum/graph/encoding/dot"
 	"gonum.org/v1/gonum/graph/multi"
+	"gonum.org/v1/gonum/graph/topo"
 )
 
 var ErrBuildingGraph = errors.New("cannot build graph")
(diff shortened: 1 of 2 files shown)


rhamzeh:
@adriantam - can we add tests for this based on our existing Cycle detection checks? https://github.com/openfga/language/blob/main/tests/data/dsl-semantic-validation-cases.yaml#L1417-L1654

pkg/go/graph/graph.go
```diff
@@ -39,4 +41,47 @@ func (g *AuthorizationModelGraph) GetDOT() string {
 	return string(dotRepresentation)
 }
 
-// TODO add graph traversals, cycle detection, etc.
+// CycleInformation encapsulates whether the graph has cycles.
+type CycleInformation struct {
+	hasCyclesAtCompileTime bool
+	hasCyclesAtRuntime     bool
```
miparnisari, line 47: can you please add some godocs here? E.g.

```
hasCyclesAtCompileTime: these are cycles that should be blocked at time of writing the model because trying to perform a Check on it will cause a stack overflow no matter what the tuples are.
canHaveCyclesAtRuntime: these are cycles that could exist if there are tuples that introduce them
```

pkg/go/graph/graph.go
```diff
@@ -39,4 +41,47 @@ func (g *AuthorizationModelGraph) GetDOT() string {
...
+}
+
+func (g *AuthorizationModelGraph) nodeListHasNonComputedEdge(nodeList []graph.Node) bool {
+	for i, nodeI := range nodeList {
+		for _, nodeJ := range nodeList[i+1:] {
+			otherEdges := g.Lines(nodeI.ID(), nodeJ.ID())
```
miparnisari, line 53: I'm curious why you have a double for loop? Can we not do this like this?

```
for i := 0; i < len(nodeList) - 1; i++ {
	this := nodeList[i]
	next := nodeList[i+1]
        // ... 
}

File: pkg/go/graph/graph.go [hunk 3]
```diff
@@ -39,4 +41,47 @@ func (g *AuthorizationModelGraph) GetDOT() string {
...
+			for otherEdges.Next() {
```
miparnisari, line 54: Nit: why `otherEdges`? I think `allEdges` is better

pkg/go/graph/graph.go
adriantam: It does not cover all potential paths - the node list only tells you what nodes are involved in the loop.  
Assume nodes 1,2,3,4 are returned.
There can be paths between Node1 and Node3 (or node 2 and node 4).  We will need to examine whether these paths are computed or not too.

adriantam:
> @adriantam - can we add tests for this based on our existing Cycle detection checks? https://github.com/openfga/language/blob/main/tests/data/dsl-semantic-validation-cases.yaml#L1417-L1654

All cases are added via test case `computed_relation_with_size_two`,  `computed_relation_with_cycle`,  `unionRuntimeRecursive`, `potential_cycle_or_but_not`, `potential_cycle_four_union`, `potential_cycle_four_union_with_one_member_no_union`

rhamzeh (approved):

pkg/go/graph/graph.go
miparnisari: Ohhh OK!

miparnisari (approved):

rhamzeh (approved):

PR #345, opened 2024-09-13
feat: add Reversed method to graph

miparnisari:
## Description

Add ability to visualize and represent graphs like this (the direction of a Check API request)

image

in addition to this (the direction of a ListObjects API request)

image

### DIFF
--- a/pkg/go/graph/graph_test.go
+++ b/pkg/go/graph/graph_test.go
@@ -1,0 +1,87 @@
+package graph
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/stretchr/testify/require"
+
+	language "github.com/openfga/language/pkg/go/transformer"
+)
+
+func TestReverseGraph(t *testing.T) {
+	t.Parallel()
+	testCases := map[string]struct {
+		model          string
+		expectedOutput string // can visualize in https://dreampuf.github.io/GraphvizOnline
+	}{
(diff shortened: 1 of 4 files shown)


pkg/go/graph/graph_test.go
```diff
@@ -0,0 +1,87 @@
...
+						define owner: [user]
+						define approved_member: member or owner
+				type group
+					relations
+						define approved_member: [user]
+				type license
+					relations
+						define active_member: approved_member from owner
+						define owner: [company, group]`,
+			expectedOutput: `digraph {
```
adriantam, line 35: Does it make sense to have a compare of with and without Reverse - we don't have the dot diagram in the unit test for the above model without reverse

rhamzeh: Maybe we can also have `authorization-model.reverse.dot` for tests, same as we have 
https://github.com/openfga/language/blob/main/tests/data/transformer/02-type-single-relation/authorization-model.dot ?

miparnisari: I thought about that, but it's soooo much work and i'm lazy 😆  would you be OK if in the existing tests i assert that `dot(reverse(reverse(graph)) == dot` ?

miparnisari: Updated @rhamzeh @adriantam

adriantam (approved):

PR #351, opened 2024-09-18
feat: assign weights in graph

miparnisari:
## Description

This is the scaffolding code required to compute weights for each node and edge. It's not meant to support every kind of model - more followup PRs will come. (For example, models with cycles are not handled).

Please review each model in the `weighted_graph_builder_test.go` and visualize the output in https://dreampuf.github.io/GraphvizOnline/. This will help you review the correctness of the weight assignment algorithm.

## Example

image

### DIFF
--- a/pkg/go/graph/weighted_graph.go
+++ b/pkg/go/graph/weighted_graph.go
@@ -14,8 +14,28 @@ type WeightedAuthorizationModelGraph struct {
 	drawingDirection DrawingDirection
 }
 
-// nolint: cyclop
+func (wg *WeightedAuthorizationModelGraph) GetDrawingDirection() DrawingDirection {
+	return wg.drawingDirection
+}
+
 func NewWeightedAuthorizationModelGraph(model *openfgav1.AuthorizationModel) (*WeightedAuthorizationModelGraph, error) {
+	weightedGraphBuilder, err := NewWeightedAuthorizationModelGraphBuilder(model)
+	if err != nil {
+		return nil, err
+	}
+
+	multigraph, ok := weightedGraphBuilder.DirectedMultigraphBuilder.(*multi.DirectedGraph)
+	if !ok {
+		return nil, fmt.Errorf("%w: could not cast to DirectedGraph", ErrBuildingGraph)
+	}
+
+	wg := &WeightedAuthorizationModelGraph{multigraph, weightedGraphBuilder.drawingDirection}
+
+	return wg, nil
+}
+
+//nolint: cyclop
+func NewWeightedAuthorizationModelGraphBuilder(model *openfgav1.AuthorizationModel) (*WeightedAuthorizationModelGraphBuilder, error) {
 	g, err := NewAuthorizationModelGraph(model)
 	if err != nil {
 		return nil, err
@@ -25,7 +45,7 @@ func NewWeightedAuthorizationModelGraph(model *openfgav1.AuthorizationModel) (*W
 		return nil, err
 	}
 
-	graphBuilder := &WeightedAuthorizationModelGraphBuilder{multi.NewDirectedGraph()}
+	graphBuilder := &WeightedAuthorizationModelGraphBuilder{multi.NewDirectedGraph(), DrawingDirectionCheck}
 
 	// Add all nodes
 	iterNodes := g.Nodes()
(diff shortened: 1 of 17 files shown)


pkg/go/graph/weighted_graph_builder_test.go
```diff
@@ -0,0 +1,384 @@
...
+
+// Node definitions.
+0 [label=folder];
+1 [label="folder#viewer - weights:[user=1]"];
+2 [label=user];
+3 [label="user:*"];
+
+// Edge definitions.
+1 -> 2 [label="direct - weights:[user=1]"];
+1 -> 3 [label=direct];
```
adriantam, line 91: Is there any weight to this edge?

pkg/go/graph/weighted_graph_builder.go
```diff
@@ -0,0 +1,168 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+//nolint:gocognit,cyclop
+func (wb *WeightedAuthorizationModelGraphBuilder) dfsToAssignWeights(curNode *WeightedAuthorizationModelNode, seen map[int64]struct{}) error {
```
adriantam, line 56: The complexity of this method is rather high.  Would it make sense to break this into smaller functions (perhaps one per major step - i.e., dfsAssignWeightsToNeighorEdges, dfsAssignWeightsToNested, etc.).  This also allows unit testing at a per method level.

adriantam (commented):
May not be for this PR.  But it will be good idea to add unit test in some of the major method as there may be cases not covered in the TestWeightedGraphBuilder test

elbuo8:
Why not use gonum's weighted graph? 🤔

pkg/go/graph/weighted_graph.go
```diff
@@ -0,0 +1,107 @@
...
+)
+
+type WeightedAuthorizationModelGraph struct {
+	*multi.DirectedGraph
+	drawingDirection DrawingDirection
+}
+
+//nolint: cyclop
+func NewWeightedAuthorizationModelGraph(model *openfgav1.AuthorizationModel) (*WeightedAuthorizationModelGraph, error) {
+	g, err := NewAuthorizationModelGraph(model)
```
elbuo8, line 19: Why not use the existing builder?

miparnisari:
> Why not use gonum's weighted graph? 🤔

@elbuo8 several reasons:
1. gonum's weighted graph allows natively assigning weights to edges _only_
2. those weights are float64 _only_

pkg/go/graph/weighted_graph_builder_test.go
miparnisari: No, `type:*` nodes don't have any weights, therefore any edges in their direction don't have weights either

pkg/go/graph/weighted_graph.go
miparnisari: The existing builder's job is to figure out the edges, their directions, etc. This builder's job is to assign weights. Since both are fairly complex jobs i'd like to keep them separate. Makes testing easier and less code per file

senojj:
The relationship between the weighted graph builder and the weighted graph is a bit odd. Currently the weighted graph has a dependency of its builder, which seems backwards.

Instead of coupling it all together with the current approach:
```go
weightedGraph, err := NewWeightedAuthorizationModelGraph(model)
```
I would suggest decoupling them explicitly:
```go
weightedGraphBuilder := NewWeightedAuthorizationModelGraphBuilder()
weightedGraph, err := weightedGraphBuilder.Build(model)
```
I know this is introducing an additional step/line of code, however I really don't see the point of having the builder, otherwise.

pkg/go/graph/weighted_graph_edge.go
```diff
@@ -0,0 +1,75 @@
...
+
+	"gonum.org/v1/gonum/graph/encoding"
+)
+
+type WeightedAuthorizationModelEdge struct {
+	*AuthorizationModelEdge
+	weights WeightMap
+
+	// isNested signals that this edge is a self-loop. The associated node will also have this flag set to true.
+	isNested bool
```
senojj, line 15: Put in other words, is this an edge that connects a vertex to itself?
```mermaid
  graph TD;
      A-->A;
```

miparnisari: Yep! not sure what to call it

PR #353, opened 2024-09-20
chore: add graph docs

miparnisari:


adriantam (approved):

pkg/go/graph/graph.go
```diff
@@ -16,8 +16,10 @@ var ErrBuildingGraph = errors.New("cannot build graph")
 type DrawingDirection bool
 
 const (
+	// DrawingDirectionListObjects is when terminal types have outgoing edges and no incoming edges.
 	DrawingDirectionListObjects DrawingDirection = true
-	DrawingDirectionCheck       DrawingDirection = false
+	// DrawingDirectionCheck is when terminal types have incoming edges and no outgoing edges.
+	DrawingDirectionCheck DrawingDirection = false
 )
```
adriantam, line 23: drawingDirection is a private member.  Can we expose this via public method?

miparnisari: Sorry I don't understand, you mean you can't access the enum from the outside??

miparnisari: Done in https://github.com/openfga/language/pull/354

PR #354, opened 2024-09-20
feat: expose graph methods

miparnisari:
## Description

Add helper methods as required by https://github.com/openfga/openfga/pull/1946

### DIFF
--- a/pkg/go/graph/graph.go
+++ b/pkg/go/graph/graph.go
@@ -25,8 +25,12 @@ const (
 
 type AuthorizationModelGraph struct {
 	*multi.DirectedGraph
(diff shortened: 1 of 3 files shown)


pkg/go/graph/graph.go
```diff
@@ -22,7 +25,28 @@ const (
 
 type AuthorizationModelGraph struct {
 	*multi.DirectedGraph
-	drawingDirection DrawingDirection
+	DrawingDirection DrawingDirection
```
miparnisari, line 28: also exposing this field

adriantam (approved):

Issue #384, opened 2024-11-19
Crash if a model is sent to TransformModuleFilesToModel


fsedano:
If a model file is sent to TransformModuleFilesToModel function, there's a null pointer dereference in transformer/module-to-model.go:119, causing app to crash:

Minimal example code:

```
package main

import (
	"log"

	language "github.com/openfga/language/pkg/go/transformer"
)

var dsl = `
model
  schema 1.1

type user
`

func main() {

	modFile := []language.ModuleFile{
		{
			Contents: dsl,
			Name:     "a",
		},
	}
	_, err := language.TransformModuleFilesToModel(modFile, "1.2")
	if err != nil {
		log.Printf("transform err: %s", err)
	}
}
```

Result:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x94d611]

goroutine 1 [running]:
github.com/openfga/language/pkg/go/transformer.TransformModuleFilesToModel({0xc000575f20, 0x1, 0x410990?}, {0xaacb78, 0x3})
        /go/pkg/mod/github.com/openfga/language/pkg/go@v0.2.0-beta.0/transformer/module-to-model.go:119 +0x1971
main.main()
        /laas/cmd/test.go:24 +0x85
exit status 2
root@polsync-okteto-664dffc4fd-g8269:/laas# 

```

ashok-an:
Could it be related to [stack overflow query](https://stackoverflow.com/questions/75893139/antlr4-4-12-0-and-go-target-visitor)? if it is then there is an open issue with antlrv4 and a work-around mentioned [in this discussion](https://github.com/antlr/antlr4/issues/2504#issuecomment-1299123230)

PR #386, opened 2024-11-20
fix: resolve crash in `TransformModuleFilesToModel` if regular schema file passed instead

fsedano:
## Description

The change checks the metadata for the module, if it's null it returns the proper error.

## References

Fixes https://github.com/openfga/language/issues/384

## Review Checklist
- [X] I have clicked on ["allow edits by maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).
- [X] I have added documentation for new/changed functionality in this PR or in a PR to [openfga.dev](https://github.com/openfga/openfga.dev) [Provide a link to any relevant PRs in the references section above]
- [X] The correct base branch is being used, if not `main`
- [X] I have added tests to validate that the change in functionality is working as expected

### DIFF
--- a/pkg/go/transformer/module-to-model.go
+++ b/pkg/go/transformer/module-to-model.go
@@ -116,8 +116,15 @@ func TransformModuleFilesToModel( //nolint:funlen,gocognit,cyclop
 			}
 
 			types = append(types, typeDef.GetType())
-			typeDef.Metadata.SourceInfo = &openfgav1.SourceInfo{
-				File: module.Name,
+			if typeDef.Metadata != nil {
+				typeDef.Metadata.SourceInfo = &openfgav1.SourceInfo{
+					File: module.Name,
+				}
+			} else {
+				transformErrors = multierror.Append(transformErrors, &ModuleTransformationSingleError{
+					Msg: "file is not a module",
+				})
+				continue
 			}
 			rawTypeDefs = append(rawTypeDefs, typeDef)
 		}
--- a/pkg/js/transformer/modules/modules-to-model.ts
+++ b/pkg/js/transformer/modules/modules-to-model.ts
@@ -49,7 +49,17 @@ export const transformModuleFilesToModel = (
           extendedTypeDefs[name].push(typeDef);
           continue;
         }
-
+        if (!typeDef.metadata) {
+          errors.push(
+            new ModuleTransformationSingleError({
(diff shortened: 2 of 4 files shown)


fsedano:
@elbuo8 - Hi, would it be possible to have a review on this? It's impacting our production code and I rather not deploy a patched version, so would love having this fixed upstream. Thanks!

rhamzeh:
@fsedano - do you require an error? What do you think about returning a transformation with the module name empty?

fsedano:
@rhamzeh In general I'd prefer to have an error, since this is most probably a input error. If the user wants to accept both it could catch the error and use the regular API for that. If we just hide the error it will go unnoticed and it might have unexpected side effects.

The way this was caught in prod is because somebody by mistake commited the wrong file as a module. I rather error out to point the user about the mistake.

rhamzeh:
> @rhamzeh In general I'd prefer to have an error, since this is most probably a input error. If the user wants to accept both it could catch the error and use the regular API for that. If we just hide the error it will go unnoticed and it might have unexpected side effects.
> 
> The way this was caught in prod is because somebody by mistake commited the wrong file as a module. I rather error out to point the user about the mistake.
> 

Thanks for elaborating, will merge this to unblock you, and in the future we'll add options to allow this use-case.

The one I currently forsee is allowing folks to run and test and transform their module without worrying about the module file and then a separate platform process does the combining - but as that is not a current use-case, it shouldn't be a blocker.

Will add an issue for us to port this change to JS and Java to maintain consistency.

May I ask how are you consuming this? Is it by calling it directly or through the CLI? (to see if we need to spin up a new CLI release that contains this fix to unblock you)

rhamzeh (approved):

fsedano:
Thanks @rhamzeh  - Just fixed also the JS one, thanks for letting me know.
We are not using CLI, but the Go SDK, so no CLI hotfix is needed.

Thanks again!

rhamzeh (approved):

fsedano:
Thanks @rhamzeh  - Is there anything else I have to do to have this merged or will it happen automagically? Thanks,

rhamzeh:
Was waiting for all the tests to pass, just added it to the queue.

Thanks for the fix!

PR #410, opened 2025-01-20
chore: add PathExists to graph

miparnisari:
## Description

The goal of this PR is to have `openfga/openga` eliminate its dependency to `gonum`. (https://github.com/openfga/openfga/blob/6d2f0703ae89caa4c94dd7678df73590edc9bfe8/go.mod#L50)

Therefore, I am moving most of this code and its tests https://github.com/openfga/openfga/blob/6d2f0703ae89caa4c94dd7678df73590edc9bfe8/pkg/typesystem/typesystem.go#L737 to this repository.

## References

https://github.com/openfga/openfga/pull/2104/


pkg/go/graph/graph.go
```diff
@@ -101,6 +101,23 @@ func (g *AuthorizationModelGraph) Reversed() (*AuthorizationModelGraph, error) {
 	return nil, fmt.Errorf("%w: could not cast to directed graph", ErrBuildingGraph)
 }
 
+// PathExists returns true if both nodes exist and there is a path starting at 'fromLabel' extending to 'toLabel'.
+// If both labels are equal, it returns true.
```
miparnisari, line 105: https://github.com/gonum/gonum/issues/2014

adriantam: Do we also want to add note here or as FIXME/TODO that for intersection and exclusion, it will not be smart enough to return false if only one side is satisified?

miparnisari: I don't think that logic or that TODO belongs here. As far as path traversal goes, our current code is correct. The logic you are talking about is more of business/app logic that belongs in the server, IMO.

adriantam (approved):

PR #441, opened 2025-04-01
removing dependency on drawing graph

yissellokta:
The weighted graph was built on top of the AuthorizationModelGraph. the AuthorizationModelGraph does not respect the order in the edges, causing problems specifically in exclusion there is no clear distinction between the left and right side

## Description
This PR will remove the dependency of building the weighted graph using the AuthorizationModelGraph instead will translate directly from the Model to the weighted graph. It also returns errors when models with invalid configurations like relations that do not exist in the type are allowed in the type system. 
All the tests used in the AuthorizationModelGraphBuilder tests were translated into the weighted graph builder tests, including ordering of the but not use case. 
It will remove the overhead of having two graphs, the weighted graph is our custom graph implementation and we can modify it to address any further needs for our algorithms, including if we need bidirectional edges. 

## Review Checklist


pkg/go/graph/weighted_graph_builder.go
```diff
@@ -18,60 +21,182 @@ func NewWeightedAuthorizationModelGraphBuilder() *WeightedAuthorizationModelGrap
...
+	computedRelation := rewrite.GetComputedUserset().GetRelation()
+
+	// find all the directly related types to the tupleset
+	relationMetadata, ok := typeDef.GetMetadata().GetRelations()[tuplesetRelation]
+	if !ok {
+		return fmt.Errorf("%w: Model cannot be parsed. %s invalid tupleset relation", ErrInvalidModel, tuplesetRelation)
 	}
+	directlyRelated := relationMetadata.GetDirectlyRelatedUserTypes()
+
+	validTTU := false
```
adriantam, line 123: instead of keep setting validTTU, would it make sense to use the following
```
if len(directlyRelated) == 0 {
    return fmt.Errorf("%w: Model cannot be parsed. No type and relation link exists for tupleset relation %s and computed relation %s", ErrInvalidModel, tuplesetRelation, computedRelation)
}
```

pkg/go/graph/weighted_graph_builder.go
```diff
@@ -18,60 +21,182 @@ func NewWeightedAuthorizationModelGraphBuilder() *WeightedAuthorizationModelGrap
...
-	err = wb.AssignWeights()
-	if err != nil {
-		return nil, err
+func (wgb *WeightedAuthorizationModelGraphBuilder) parseComputed(wg *WeightedAuthorizationModelGraph, parentNode *WeightedAuthorizationModelNode, typeDef *openfgav1.TypeDefinition, relation string) {
+	nodeType := RewriteEdge
+	// e.g. define x: y. Here y is the rewritten relation
+	rewrittenNodeName := fmt.Sprintf("%s#%s", typeDef.GetType(), relation)
+	newNode := wg.GetOrAddNode(rewrittenNodeName, rewrittenNodeName, SpecificTypeAndRelation)
+	// new edge from x to y
+	if parentNode.nodeType == SpecificTypeAndRelation && newNode.nodeType == SpecificTypeAndRelation {
```
adriantam, line 164: can you describe when this will not be true?

pkg/go/graph/weighted_graph.go
```diff
@@ -57,6 +57,19 @@ func (wg *WeightedAuthorizationModelGraph) AddNode(uniqueLabel, label string, no
 	wg.nodes[uniqueLabel] = &WeightedAuthorizationModelNode{uniqueLabel: uniqueLabel, label: label, nodeType: nodeType, wildcards: wildcards}
 }
 
+// AddNode adds a node to the graph with optional nodeType and weight.
+func (wg *WeightedAuthorizationModelGraph) GetOrAddNode(uniqueLabel, label string, nodeType NodeType) *WeightedAuthorizationModelNode {
```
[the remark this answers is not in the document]
yissellokta: a wildcard node is a different node than a non wildcard node. the unique label which identified the node is different it will be like creating a node named "user" and a node named "employee". For the user type, if we have a wildcard user reference then the node will be built as "user:*" , if it is none wildcard then will be built as "user"

pkg/go/graph/weighted_graph.go
```diff
@@ -57,6 +57,19 @@ func (wg *WeightedAuthorizationModelGraph) AddNode(uniqueLabel, label string, no
...
+	if existingNode := wg.nodes[uniqueLabel]; existingNode != nil {
+		return existingNode
+	}
+	wildcards := make([]string, 0)
+	if nodeType == SpecificTypeWildcard {
+		wildcards = append(wildcards, uniqueLabel[:len(uniqueLabel)-2])
+	}
+	wg.nodes[uniqueLabel] = &WeightedAuthorizationModelNode{uniqueLabel: uniqueLabel, label: label, nodeType: nodeType, wildcards: wildcards}
```
senojj, line 69: Looking at how the `wildcards` property is used throughout the `WeightedAuthorizationModelGraph`, you can avoid many unnecessary allocations by leaving `wildcards` equal to `nil` when there are no wildcard values present.
```suggestion
	var wildcards []string
	if nodeType == SpecificTypeWildcard {
		wildcards = append(wildcards, uniqueLabel[:len(uniqueLabel)-2])
	}
	wg.nodes[uniqueLabel] = &WeightedAuthorizationModelNode{uniqueLabel: uniqueLabel, label: label, nodeType: nodeType, wildcards: wildcards}
```

pkg/go/graph/weighted_graph_builder.go
yissellokta: when you are going from an Operator node to another relation

pkg/go/graph/weighted_graph.go
```diff
@@ -68,6 +81,54 @@ func (wg *WeightedAuthorizationModelGraph) AddEdge(fromID, toID string, edgeType
...
+			if edge.edgeType == edgeType && edge.tuplesetRelation == tuplesetRelation {
+				for _, cond := range edge.conditions {
+					if cond == condition {
+						return nil
+					}
+				}
+				edge.conditions = append(edge.conditions, condition)
+				return nil
+			}
+		}
```
senojj, line 105: cognitive complexity: can reduce one level of nesting
```suggestion
		if edge.to.uniqueLabel == toNode.uniqueLabel && edge.edgeType == edgeType && edge.tuplesetRelation == tuplesetRelation {
			for _, cond := range edge.conditions {
				if cond == condition {
					return nil
				}
			}
			edge.conditions = append(edge.conditions, condition)
			return nil
		}
```

pkg/go/graph/weighted_graph.go
```diff
@@ -68,6 +81,54 @@ func (wg *WeightedAuthorizationModelGraph) AddEdge(fromID, toID string, edgeType
...
+	}
+
+	conditions := []string{condition}
+	wildcards := make([]string, 0)
+
+	edge := &WeightedAuthorizationModelEdge{from: fromNode, to: toNode, edgeType: edgeType, tuplesetRelation: tuplesetRelation, wildcards: wildcards, conditions: conditions}
```
senojj, line 111: reduce allocations: wildcards can be left `nil` here.
```suggestion
	edge := &WeightedAuthorizationModelEdge{from: fromNode, to: toNode, edgeType: edgeType, tuplesetRelation: tuplesetRelation, conditions: conditions}
```

pkg/go/graph/weighted_graph.go
```diff
@@ -68,6 +81,54 @@ func (wg *WeightedAuthorizationModelGraph) AddEdge(fromID, toID string, edgeType
...
+	if fromNode == nil || toNode == nil {
+		return false
+	}
+	edges := wg.edges[fromNode.uniqueLabel]
+	for _, edge := range edges {
+		if edge.to.uniqueLabel == toNode.uniqueLabel {
+			if edge.edgeType == edgeType && edge.tuplesetRelation == tuplesetRelation {
+				return true
+			}
+		}
```
senojj, line 126: cognitive complexity: can reduce nesting by one level.
```suggestion
		if edge.to.uniqueLabel == toNode.uniqueLabel && edge.edgeType == edgeType && edge.tuplesetRelation == tuplesetRelation {
			return true
		}
```

pkg/go/graph/weighted_graph_builder.go
```diff
@@ -18,60 +21,180 @@ func NewWeightedAuthorizationModelGraphBuilder() *WeightedAuthorizationModelGrap
...
+		// sort relations by name to guarantee stable output
+		sortedRelations := make([]string, 0, len(typeDef.GetRelations()))
+		for relationName := range typeDef.GetRelations() {
+			sortedRelations = append(sortedRelations, relationName)
+		}
+
+		slices.Sort(sortedRelations)
+
+		for _, relation := range sortedRelations {
+			uniqueLabel := fmt.Sprintf("%s#%s", typeDef.GetType(), relation)
```
senojj, line 45: performance: direct concatenation will be far faster than using `fmt.Sprintf`
```suggestion
			uniqueLabel := typeDef.GetType() + '#' + relation
```

pkg/go/graph/weighted_graph_builder.go
```diff
@@ -18,60 +21,180 @@ func NewWeightedAuthorizationModelGraphBuilder() *WeightedAuthorizationModelGrap
...
+	}
+
+	operatorNodeName := fmt.Sprintf("%s:%s", operator, ulid.Make().String())
+	operatorNode := wg.GetOrAddNode(operatorNodeName, operator, OperatorNode)
+
+	// add one edge "relation" -> "operation that defined the operator"
+	// Note: if this is a composition of operators, operationNode will be nil and this edge won't be added.
+	wg.AddEdge(parentNode.GetUniqueLabel(), operatorNodeName, RewriteEdge, "", nil)
+	for _, child := range children {
+		err := wgb.parseRewrite(wg, operatorNode, model, child, typeDef, relation)
```
senojj, line 99: given a specific model, this recursion could overflow the stack. I would suggest bounding the recursion, or using a `for` loop instead.

pkg/go/graph/weighted_graph_builder.go
```diff
@@ -18,60 +21,180 @@ func NewWeightedAuthorizationModelGraphBuilder() *WeightedAuthorizationModelGrap
...
-			}
-			castedToNode, ok := castedEdge.To().(*AuthorizationModelNode)
-			if !ok {
-				return nil, fmt.Errorf("%w: could not cast %v to AuthorizationModelNode", ErrBuildingGraph, castedEdge.To())
-			}
-			wb.AddEdge(castedFromNode.uniqueLabel, castedToNode.uniqueLabel, castedEdge.edgeType, castedEdge.tuplesetRelation, castedEdge.conditions)
+		validTTU = true
+		rewrittenNodeName := fmt.Sprintf("%s#%s", tuplesetType, computedRelation)
+		nodeSource := wg.GetOrAddNode(rewrittenNodeName, rewrittenNodeName, SpecificTypeAndRelation)
+		typeTuplesetRelation := fmt.Sprintf("%s#%s", typeDef.GetType(), tuplesetRelation)
```
senojj, line 132: performance: direct concatenation will be much faster than fmt.Sprintf
```suggestion
		typeTuplesetRelation := typeDef.GetType() + "#" + tuplesetRelation
```

pkg/go/graph/weighted_graph_builder.go
```diff
@@ -18,60 +21,180 @@ func NewWeightedAuthorizationModelGraphBuilder() *WeightedAuthorizationModelGrap
...
+	return nil
+}
 
-	err = wb.AssignWeights()
-	if err != nil {
-		return nil, err
+func (wgb *WeightedAuthorizationModelGraphBuilder) parseComputed(wg *WeightedAuthorizationModelGraph, parentNode *WeightedAuthorizationModelNode, typeDef *openfgav1.TypeDefinition, relation string) {
+	nodeType := RewriteEdge
+	// e.g. define x: y. Here y is the rewritten relation
+	rewrittenNodeName := fmt.Sprintf("%s#%s", typeDef.GetType(), relation)
```
senojj, line 159: performance: direct concatenation will be much faster than fmt.Sprintf
```suggestion
	rewrittenNodeName := typeDef.GetType() + "#" + relation
```

pkg/go/graph/weighted_graph.go
adriantam: Actually, thinking it a little bit more, we don't need the append at all as we already know that wildcards is an empty array
```
		var wildcards []string
		if nodeType == SpecificTypeWildcard {
			wildcards = []string{ uniqueLabel[:len(uniqueLabel)-2]}
		}
		wg.nodes[uniqueLabel] = &WeightedAuthorizationModelNode{uniqueLabel: uniqueLabel, label: label, nodeType: nodeType, wildcards: wildcards}
```

pkg/go/graph/weighted_graph_builder_test.go
```diff
@@ -600,3 +600,952 @@ func TestValidConditionalGraphModel(t *testing.T) {
...
+	graph, err := wgb.Build(authorizationModel)
+	require.NoError(t, err)
+
+	require.Len(t, graph.nodes, 10)
+	require.Len(t, graph.edges, 5)
+	exclusionNodeID := graph.edges["job#can_read"][0].to.uniqueLabel
+	require.Len(t, graph.edges[exclusionNodeID], 5)
+	cannotreadID := graph.edges[exclusionNodeID][4].to.uniqueLabel
+	require.Equal(t, "job#cannot_read", cannotreadID)
+}
```
adriantam, line 633: would it be beneficial to assert the order of the can_read edge as this is one of the reason why we need to remove the dependency (i.e., to guarantee order)

pkg/go/graph/weighted_graph_builder_test.go
```diff
@@ -600,3 +600,952 @@ func TestValidConditionalGraphModel(t *testing.T) {
...
+                type user
+	`
+	authorizationModel := language.MustTransformDSLToProto(model)
+	wgb := NewWeightedAuthorizationModelGraphBuilder()
+	graph, err := wgb.Build(authorizationModel)
+	require.NoError(t, err)
+
+	require.Len(t, graph.nodes, 3)
+	require.Len(t, graph.edges, 1)
+	require.True(t, graph.nodes["user"].nodeType == SpecificType)
```
adriantam, line 652: Comments to apply other similar pattern.  Use `require.Equal(t, y, x)` instead of `require.True(t, x == y)` as it give better error message when not equal.

```suggestion
	require.Equal(t, SpecificType, graph.nodes["user"].nodeType)
```

adriantam (changes requested):

pkg/go/graph/weighted_graph.go
senojj: Using append, or creating the array outright are both acceptable here; I'm not certain either would yield a significant performance benefit in this particular case.

adriantam (approved):

PR #443, opened 2025-04-07
fix: exclusion weight where type is not in the excluded relation

adriantam:
## Description
For relation A but not B, we need to use a mixed strategy in calculating weight where a type in A will use the max strategy and B will use the enforced type strategy. This is because a type can be connected to the graph if it appears only in A as B is not mandatory.  However, a type listed in B MUST also appear in A.  Otherwise, relation will never exist for such type.

## Review Checklist

### DIFF
--- a/pkg/go/graph/weighted_graph.go
+++ b/pkg/go/graph/weighted_graph.go
@@ -356,20 +356,41 @@ func (wg *WeightedAuthorizationModelGraph) calculateNodeWeightFromTheEdges(nodeI
 	var err error
 	node := wg.nodes[nodeID]
 
-	// if there is not cycle, we can calculate the weight without taking in consideration the cycle complexity.
+	// if there is not a cycle, we can calculate the weight without taking in consideration the cycle complexity.
 	if len(tupleCycles) == 0 {
-		// for any node that is not AND or BUT NOT, we can calculate the weight using the max strategy
-		if node.nodeType != OperatorNode || node.label == UnionOperator {
-			err = wg.calculateNodeWeightWithMaxStrategy(nodeID)
+		// for any non-operator node, we can calculate the weight using the max strategy because
+		// a type is valid as long as it appears in one of the branch.
+		if node.nodeType != OperatorNode {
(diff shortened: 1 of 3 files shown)


pkg/go/graph/weighted_graph.go
```diff
@@ -442,7 +463,37 @@ func (wg *WeightedAuthorizationModelGraph) calculateNodeWeightWithMaxStrategy(no
...
+	if len(edges) == 0 && node.nodeType != SpecificType && node.nodeType != SpecificTypeWildcard {
+		return fmt.Errorf("%w: %s node does not have any terminal type to reach to", ErrInvalidModel, node.uniqueLabel)
+	}
+
+	for idx, edge := range edges {
+		for key, value := range edge.weights {
+			if _, ok := weights[key]; !ok {
+				if idx != len(edges)-1 {
+					// This is the A edge.  We take the max weight of all key
+					weights[key] = value
```
yissellokta, line 484: you still need to take the max value not the last value, 
For example in the case I have rel1: [user, employee, type#rel2, type#rel5] but not rel6 
if type#rel2 has a weight of user to 4 and if type#rel5 has a weight of user to be 2,
the weight for the left side should be 4 and not 2 for user

yissellokta (changes requested):

pkg/go/graph/weighted_graph.go
yissellokta: Let's add a test that covers this use case too

adriantam: When we are in line 484, `weights[key]` is not found because of line 481.  If it is found (i.e., have something to get max), it will be handled in line 487

adriantam: In terms of test, what you are describing should be covered in B_appears_in_A_infinite (line 513 of [weighted_graph_builder_test.go](https://github.com/openfga/language/pull/443/files/3d95b794befc8f582159837f0f641c32c0733e22#diff-ed83f977a716c3831d01dc92123b06eb123fe64fe4154288d62f079f3fe4c670))

yissellokta (approved):

PR #455, opened 2025-05-28
Fix: correctly assign weights for intersections with multiple direct types

willvedd:
## Description
Currently, there is a bug when assigning weights to intersection operations involving more than two direct terminal types. Two examples illustrate this issue:

First, the `user3` type should be pruned from the edge because that type assignment does not exist on both sides of the `and` operation.
Screenshot 2025-05-28 at 10 36 46 AM

Secondly, and more serious, the types on both sides of the `and` _are_ balanced but only the `user3` weight exists on the edge.
Screenshot 2025-05-28 at 10 36 35 AM

**NOTE:** This changes causes the `TestInvalidWeight1NotMatchingTerminalTypes` test to fail but it begs the question about what the correct behavior is when types are unbalanced across the `and` operation; is it better to prohibit the model or quietly prune away the weights?

## Review Checklist

## Summary by CodeRabbit

- **Tests**
  - Improved clarity of test assertions for graph structure and edge types.
  - Expanded test coverage with new scenarios for intersection relations, including multiple operands and mixed assignment types.
  - Enhanced formatting and readability of test cases.


pkg/go/graph/weighted_graph.go
```diff
@@ -499,32 +499,35 @@ func (wg *WeightedAuthorizationModelGraph) calculateNodeWeightWithMixedStrategy(
...
 
-		// for AndOperation, remove the key if it is not in the edge, not all edges return the same type
-		for key := range weights {
-			if value, ok := edge.weights[key]; !ok {
-				delete(weights, key)
+	for i, edge := range edges {
+		isLastElement := len(edges) == i+1
+
+		for key, value := range edge.weights {
+			if isLastElement {
```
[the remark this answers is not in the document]
willvedd: Yes, that was a fundamental misunderstanding on my part. I've now updated to account for imbalances across multiple `and` statements.

pkg/go/graph/weighted_graph_builder_test.go
```diff
@@ -1129,47 +1129,390 @@ func TestGraphConstructionInvalidTTU(t *testing.T) {
...
+			require.True(t, graph.edges[andNode.uniqueLabel][0].edgeType == RewriteEdge)
+			require.True(t, graph.edges[andNode.uniqueLabel][1].to.uniqueLabel == "folder#b")
+			require.True(t, graph.edges[andNode.uniqueLabel][1].edgeType == RewriteEdge)
+			require.Len(t, graph.edges["folder#a"], 1)
+			require.True(t, graph.edges["folder#a"][0].edgeType == DirectEdge)
+			require.Len(t, graph.edges["folder#b"], 1)
+			require.True(t, graph.edges["folder#b"][0].edgeType == DirectEdge)
+			require.Len(t, graph.edges["folder#c"], 1)
+			require.True(t, graph.edges["folder#c"][0].edgeType == RewriteEdge)
+		})
```
adriantam, line 1174: Is there assertion on weight?

pkg/go/graph/weighted_graph.go
```diff
@@ -499,32 +499,50 @@ func (wg *WeightedAuthorizationModelGraph) calculateNodeWeightWithMixedStrategy(
...
+	directlyAssignableWeights := make(map[string]int)
+	rewriteWeights := make(map[string]int)
+
 	for _, edge := range edges {
-		// for but not ensure that the first edge is the left edge
-		// the first time, take the weights of the edge
-		if len(weights) == 0 {
-			for key, value := range edge.weights {
-				weights[key] = value
+		if edge.GetEdgeType() == DirectEdge {
```
adriantam, line 512: Please add comments that we assume directEdge will always be processed before non-direct edges.

pkg/go/graph/weighted_graph.go
```diff
@@ -499,32 +499,50 @@ func (wg *WeightedAuthorizationModelGraph) calculateNodeWeightWithMixedStrategy(
...
+	directlyAssignableTypesExist := len(directlyAssignableWeights) > 0
+	if directlyAssignableTypesExist {
+		for key := range directlyAssignableWeights {
+			if _, existsInBoth := rewriteWeights[key]; existsInBoth {
+				weights[key] = int(math.Max(float64(rewriteWeights[key]), float64(directlyAssignableWeights[key])))
 			}
 		}
+	} else {
+		weights = rewriteWeights
 	}
```
adriantam, line 544: Suggestions: Instead of creating another weights map, why not have something like

```
if directlyAssignableTypesExist {
   for key := range rewriteWeights {
    item, existsInBoth := directlyAssignableWeights[key]
    if !existsInBoth {
        delete (rewriteWeights, key)
    }
    rewriteWeights[key] = int(math.Max(item,  rewriteWeights[key]))
   }
   weights := rewriteWeights
}

```
In this way, you don't allocate the map twice.

pkg/go/graph/weighted_graph.go
```diff
@@ -499,32 +499,50 @@ func (wg *WeightedAuthorizationModelGraph) calculateNodeWeightWithMixedStrategy(
...
 	node := wg.nodes[nodeID]
-	weights := make(map[string]int)
 	edges := wg.edges[nodeID]
 
 	if len(edges) == 0 && node.nodeType != SpecificType && node.nodeType != SpecificTypeWildcard {
 		return fmt.Errorf("%w: %s node does not have any terminal type to reach to", ErrInvalidModel, node.uniqueLabel)
 	}
 
+	directlyAssignableWeights := make(map[string]int)
+	rewriteWeights := make(map[string]int)
```
adriantam, line 509: Please allocate with size len(edges)

pkg/go/graph/weighted_graph_builder_test.go
```diff
@@ -1129,47 +1129,390 @@ func TestGraphConstructionInvalidTTU(t *testing.T) {
...
+				define c: a and b
+		  `
+			authorizationModel := language.MustTransformDSLToProto(model)
+			wgb := NewWeightedAuthorizationModelGraphBuilder()
+			graph, err := wgb.Build(authorizationModel)
+			require.NoError(t, err)
+
+			require.Len(t, graph.nodes, 6)
+			require.Len(t, graph.edges, 4)
+			require.True(t, graph.nodes["user"].nodeType == SpecificType)
```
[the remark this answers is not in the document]
willvedd: Can you clarify what you mean by "apply to rest"? What do you want me to apply?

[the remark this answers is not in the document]
adriantam: The unit test have many cases with similar pattern require.True(t, A == B).  We should use require.Equal(t, B, A)

pkg/go/graph/weighted_graph.go
willvedd: **Question:** Can you confirm your feedback? This code does _not_ assume any ordering. For example, if this syntax were valid we'd still have the same correct outcome:

```
relation can_view: owner and viewer and [user,employee]
```

We maintain correctness by separating the weights across two maps.

pkg/go/graph/weighted_graph.go
```diff
@@ -499,36 +499,58 @@ func (wg *WeightedAuthorizationModelGraph) calculateNodeWeightWithMixedStrategy(
...
-			} else {
-				weights[key] = int(math.Max(float64(weights[key]), float64(value)))
+	directlyAssignableTypesExist := len(directlyAssignableWeights) > 0
+	if directlyAssignableTypesExist {
+		for key := range directlyAssignableWeights {
+			if _, existsInBoth := rewriteWeights[key]; existsInBoth {
+				if node.weights == nil {
+					node.weights = make(map[string]int)
+				}
+				node.weights[key] = int(math.Max(float64(rewriteWeights[key]), float64(directlyAssignableWeights[key])))
```
willvedd, line 543: @adriantam I wasn't able to get it to work solely manipulating the `rewriteWeights` map alone but I was able to eliminate the `weights` map allocation by reusing `node.weights`.

pkg/go/graph/weighted_graph.go
```diff
@@ -499,36 +499,58 @@ func (wg *WeightedAuthorizationModelGraph) calculateNodeWeightWithMixedStrategy(
...
+	}
 
-		// for AndOperation, remove the key if it is not in the edge, not all edges return the same type
-		for key := range weights {
-			if value, ok := edge.weights[key]; !ok {
-				delete(weights, key)
-			} else {
-				weights[key] = int(math.Max(float64(weights[key]), float64(value)))
+	directlyAssignableTypesExist := len(directlyAssignableWeights) > 0
+	if directlyAssignableTypesExist {
```
adriantam, line 537:
```suggestion
	if len(directlyAssignableWeights) > 0 {
```
No need to create `directlyAssignableTypesExist`

pkg/go/graph/weighted_graph.go
```diff
@@ -499,36 +499,58 @@ func (wg *WeightedAuthorizationModelGraph) calculateNodeWeightWithMixedStrategy(
...
+		for key := range directlyAssignableWeights {
```
adriantam, line 538: Instead of checking for `node.weights == nil` inside the for loop, we can do it before to avoid un-necessary check

pkg/go/graph/weighted_graph.go
```diff
@@ -499,36 +499,58 @@ func (wg *WeightedAuthorizationModelGraph) calculateNodeWeightWithMixedStrategy(
...
+			if _, existsInBoth := rewriteWeights[key]; existsInBoth {
+				if node.weights == nil {
+					node.weights = make(map[string]int)
```
adriantam, line 541: can allocate it against the min size of len(rewriteWeights) and len(directlyAssignableWeights)

adriantam (dismissed):

pkg/go/graph/weighted_graph.go
willvedd: That would allocate for all cases but you technically only need it when this condition is satisfied:

```
if _, existsInBoth := rewriteWeights[key]; existsInBoth {
....
```

PR #624, opened 2026-07-03
feat: allow '-' and '/' in object IDs

### LINKED ISSUE #437: Update the Validation rules to account for allowing `.`, `/`
After https://github.com/openfga/language/pull/254

We'll need to do a follow-up of improving the validators:
- https://github.com/openfga/language/blob/main/pkg/js/validator/validate-rules.ts
- https://github.com/openfga/language/blob/main/pkg/go/validation/validation-rules.go
- https://github.com/openfga/language/blob/main/pkg/java/src/main/java/dev/openfga/language/validation/Validator.java

So that we can make use of these extra fields in the type in the CLI, VSCode and IntelliJ

_Originally posted by @rhamzeh in https://github.com/openfga/language/issues/254#issuecomment-2721369594_

SoulPancake:
Extend the object ID validation rule to permit '-' and '/' ('|' already allowed) across the Go, JS, and Java validators, matching characters the server accepts. Type/relation/condition rules unchanged. Closes #437.

## Review Checklist

## Summary by CodeRabbit

* **New Features**
  * Identifier validation now accepts `-` and `/` in non-leading positions across Go, Java, and JavaScript implementations.

* **Tests**
  * Added coverage for IDs containing dashes, slashes, and mixed special characters to confirm the updated validation behavior.


rhamzeh (commented):
A minor note in case you can quickly spike it - would you mind trying to see if you can move the test cases to the common yaml so that we don't have to write them 3 times and ensure they're all consistent with each other?

Not a blocker - just a quick spike

rhamzeh (approved):

PR #628, opened 2026-07-06
fix: correct error locations when type/schema lines have trailing comments

### LINKED ISSUE #393: Language is not properly exporting error locations when there are comments in the model
Image

```py
model
  schema 1.2

type group # module: common, file: common.fga
  relations
    define member: [user]

type user # module: common, file: common.fga

type page # module: confluence, file: confluence.fga
  relations
    define owner: member

type ticket # module: jira, file: jira.fga
  relations
    define reporter: member
```

SoulPancake:
The DSL validator resolved error line numbers with anchored regexes (`^type NAME$`, `^schema V$`) that failed to match lines carrying a trailing comment (e.g. `type page # module: ...`). A miss returned -1, which was then propagated as a negative `skipIndex` into the relation lookup, producing nonsensical negative lines (-1, -2) and, for the invalid-type path, a crash (`rawLine.split` on `undefined`).

Make the type/schema lookups comment-tolerant and guard against negative `skipIndex` propagation in both the JS validator and its module transformer copy, and mirror the fix in the Java validator. Add 7 shared regression cases (simple + the full reported model) covering missing-definition, reserved type/relation names, invalid type, no entrypoint, and invalid schema with trailing comments.

## Review Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation error line and column reporting when source lines include trailing comments.
  * Prevented invalid or negative line indexes from causing incorrect error locations.
  * Fixed several cases where missing schema, type, or relation errors could point to the wrong line.
* **Tests**
  * Added coverage for semantic validation scenarios involving trailing comments and module-style DSL input.


SoulPancake (commented):
TODO: After Go semantic validation is generally done, we can run these fixtures and see if it needs the same fix too

SoulPancake:
Tested it in the VSCode extension 

image
image

pkg/js/validator/validate-dsl.ts
```diff
@@ -292,47 +292,67 @@ function hasEntryPointOrLoop(
...
   if (!lines) {
     return undefined;
   }
-  return lines.slice(skipIndex).findIndex((line: string) => line.trim().match(`^type ${typeName}$`)) + skipIndex;
+  // Allow an optional trailing comment (e.g. `type page # module: ...`) after the type name.
+  // Match the type name literally (it may contain regex metacharacters like `.`).
+  const typePrefix = `type ${typeName}`;
+  const index = lines.slice(skipIndex).findIndex((line: string) => {
+    const trimmed = line.trim();
+    return trimmed.startsWith(typePrefix) && /^\s*(#.*)?$/.test(trimmed.slice(typePrefix.length));
```
rhamzeh:
```suggestion
    return trimmed.startsWith(typePrefix) && /^\s+(#.*)?$/.test(trimmed.slice(typePrefix.length));
```

require a space before the comment, otherwise you'll trigger on e.g. team#member

SoulPancake: Thanks!
I pushed a fix. `\s+` would actually regress the common no-comment case (type user → empty remainder → no match), and the `team#member` example can't reach this matcher since it only tests type declration lines. But  `\s*(#.*)?` did treat a glued # as a comment. Changed both the type and schema matchers to `/^(\s+#.*)?$/`, which requires a space before an inline comment while still allowing the empty case.

rhamzeh (approved):

PR #652, opened 2026-08-04
fix(pkg/go/utils): match declarations by full name, not by prefix

### LINKED ISSUE #255: Incorrect line for type when looking up the typeIndex, when type is a substring
**Problem**: incorrect type index return for `asset` when `asset-category` is given precedence in declaration.

Example:
image

image

**Fix**: update each method to do a full string match, not a prefix match.

https://github.com/openfga/language/blob/7d0da9bc9c63f3885c512795279f0f0a025452b1/pkg/js/util/line-numbers.ts#L23

https://github.com/openfga/language/blob/7d0da9bc9c63f3885c512795279f0f0a025452b1/pkg/go/utils/line-numbers.go#L16

https://github.com/openfga/language/blob/7d0da9bc9c63f3885c512795279f0f0a025452b1/pkg/java/src/main/java/dev/openfga/language/validation/Dsl.java#L50

SoulPancake:
The line-number helpers in `pkg/go/utils` matched a declaration with `strings.HasPrefix` on the name alone, so a name that is a prefix of another resolved to the wrong line: asking for `document` matched `type documentation`, `owner` matched `define owner_group:`, and `less` matched `condition less_than(`.

#### What problem is being solved?

`GetTypeLineNumber`, `GetRelationLineNumber`, `GetConditionLineNumber` and `GetExtendedTypeLineNumber` each tested only that a trimmed line started with `"<keyword> <name>"`, with nothing checking what followed the name:

```go
func GetTypeLineNumber(typeName string, lines []string) int {
	return slices.IndexFunc(lines, func(line string) bool {
		return strings.HasPrefix(strings.TrimSpace(line), "type "+typeName)
	})
}
```

Reproduced against `develop` — every one of these returns the line of the longer declaration that happens to come first:

| query | matched | returned | correct |
| --- | --- | --- | --- |
| type `document` | `type documentation` | 2 | 6 |
| relation `owner` | `define owner_group:` | 4 | 5 |
| condition `less` | `condition less_than(` | 7 | 8 |

The helpers have four non-test callers, all in `pkg/go/transformer/module-to-model.go`: the duplicate-type, duplicate-condition, extended-type and duplicate-relation errors. In a module set where one name is a prefix of another, those errors point at the wrong declaration.

This is user-visible in a released package. `pkg/go/utils/line-numbers.go` and `pkg/go/transformer/module-to-model.go` both ship in `pkg/go/v0.3.1`, which is the tag `openfga/cli` pins.

This is the Go half of #255, which listed all three implementations' files. #260 closed it after fixing only the JS and Java copies — `pkg/go/utils/line-numbers.go` was named in the issue but never in the diff, so the Go bug has been live since.

#### How is it being solved?

By requiring the name to *end* where the prefix ends. The four helpers now share one matcher, and the terminator rule is derived from the lexer grammar rather than hand-written per helper:

```go
// IDENTIFIER admits letters, digits, `_` and MINUS; EXTENDED_IDENTIFIER adds
// SLASH and DOT (OpenFGALexer.g4).
func IsNameByte(b byte) bool
```

`declarationIndex` finds the first line starting with the prefix whose next byte is not a name byte (or which ends there). Because the prefix always spans whole words — `type <name>`, `define <name>`, `condition <name>` — that single check is what stops `document` matching `documentation`.

Taking the terminator set from the grammar rather than requiring a specific punctuation character is what makes `type core.doc` a declaration of `core.doc` and not of `core`, since `.` is a name byte in `EXTENDED_IDENTIFIER`. A rule that instead demanded `:` or `(` after the name would get that case wrong.

Only the relation helper collapses runs of spaces, mirroring the reference implementation, where ` {2,}` normalization is applied in `getRelationLineNumber` alone — so `define  owner:` still matches `owner`.

#### What changes are made to solve it?

- `pkg/go/utils/line-numbers.go`: adds `IsNameByte` and the shared `declarationIndex`; the four exported helpers become one-line calls into it. Signatures, the `int` return and the `-1` miss sentinel are all unchanged.
- `pkg/go/validation/name_validation.go`: `GetConditionLineNumber` had its own inline terminator check requiring `(` after the name. It now calls `utils.IsNameByte`, so the rule lives in one place. This is a deliberate loosening — see the note below.
- `tests/data/transformer-module/07-prefix-collisions/`: a module fixture where `documentation`/`document` and `less`/`less_than` collide across two files, asserting the duplicate-condition error lands on `dup.fga` line 10 rather than an earlier `less_than` declaration.
- `pkg/go/utils/line-numbers_test.go`: 47 cases covering each helper, including the trailing-comment form the module fixtures depend on (`type other # module: core, file: core.fga`), collapsed spaces, dotted names, and the three collisions above.

The only visible behaviour change is that the four transformer errors point at the correct line in collision cases. No error identity or message text changes.

#### A note on what this does not tighten

The matcher accepts anything that is not a name byte as a terminator, which is looser than the punctuation each declaration form actually requires. All of these match today:

```
type doc#x            queried as `doc`     -> matches
type doc:x            queried as `doc`     -> matches
define owner [user]   queried as `owner`   -> matches   (no colon)
condition less {      queried as `less`    -> matches   (no parameter list)
```

None of these are valid DSL, and the helpers are only ever called with a name the parser has already accepted from a model that parsed — so a malformed line cannot reach them in practice. Deriving the rule from the grammar was preferred over four hand-written punctuation checks because it is the same rule in every helper and it handles extended identifiers correctly. Worth flagging explicitly rather than leaving it to be discovered.

For the same reason, the two Go copies are now closer but not identical. `pkg/go/validation`'s type and relation helpers parse the line (`strings.Fields`, then an exact compare; or split on `:`), so they reject `type doc#x`, `type doc:x` and `define owner [user]` where `utils` accepts them. Both agree on every well-formed input, including all three collision cases and the trailing-comment form. Unifying the two copies is a follow-up; this PR makes them agree on the cases that occur, not on malformed input.

#### Testing

From `pkg/go`: `go build ./...` clean, and `go test ./... -count=1` green — `utils`, `transformer` and `validation` all pass, with the new module fixture exercised through the transformer path.

The three collisions were verified as failing on `develop` and fixed on this branch by building the helpers standalone against both versions, rather than only through the test suite.

## References

closes https://github.com/openfga/language/issues/255

- #260 — closed #255 after fixing the JS and Java copies only; this is the Go half it missed
- #628 — the JS/Java trailing-comment fix, whose behaviour these helpers are matched against

## Review Checklist

### DIFF
--- a/pkg/go/utils/line-numbers.go
+++ b/pkg/go/utils/line-numbers.go
@@ -28,13 +28,12 @@ func IsNameByte(b byte) bool {
 	return nameBytes[b]
 }
 
-// declarationIndex returns the index of the first line beginning with prefix whose
-// name ends there, or -1. prefix always spans whole words (`type <name>`, `define
+// name ends there, or -1. Prefix always spans whole words (`type <name>`, `define
 // <name>`, `condition <name>`), so requiring the next byte to end the name is what
 // stops `document` matching a declaration of `documentation`.
-func declarationIndex(lines []string, prefix string, normalize func(string) string) int {
+func declarationIndex(lines []string, prefix string) int {
 	return slices.IndexFunc(lines, func(line string) bool {
-		trimmed := normalize(strings.TrimSpace(line))
+		trimmed := NormalizeWhitespace(strings.TrimSpace(line))
 		if !strings.HasPrefix(trimmed, prefix) {
 			return false
 		}
@@ -45,42 +44,78 @@ func declarationIndex(lines []string, prefix string, normalize func(string) stri
 	})
 }
 
-// keepSpaces leaves a line as-is. Only the relation helper collapses repeated
-// spaces, matching the reference, where ` {2,}` normalization is applied in
-// getRelationLineNumber alone.
-func keepSpaces(line string) string { return line }
+// isInlineWhitespace reports whether b is a byte the lexer's WHITESPACE rule
+// accepts between tokens on a line: space, tab or form feed (OpenFGALexer.g4).
+func isInlineWhitespace(b byte) bool {
+	return b == ' ' || b == '\t' || b == '\f'
+}
 
-// normalizeSpaces collapses runs of spaces into one, mirroring the reference's
-// ` {2,}` normalization, so `define  owner:` is matched the same as `define owner:`.
-func normalizeSpaces(line string) string {
-	for strings.Contains(line, "  ") {
-		line = strings.ReplaceAll(line, "  ", " ")
+// NormalizeWhitespace collapses every run of inline whitespace into one space, so
+// `define\towner:` and `define  owner:` are matched the same as `define owner:`.
+// Only space, tab and form feed are folded — exactly the lexer's WHITESPACE
+// alphabet; anything else (nbsp, vertical tab) fails to lex and can never reach a
+// line lookup. Lines that are already normal are returned unchanged.
+func NormalizeWhitespace(line string) string {
+	for i := range len(line) {
+		if b := line[i]; b == '\t' || b == '\f' || (b == ' ' && i > 0 && line[i-1] == ' ') {
+			return foldInlineWhitespace(line)
+		}
 	}
 
 	return line
 }
 
+func foldInlineWhitespace(line string) string {
+	var sb strings.Builder
+
+	sb.Grow(len(line))
+
+	inRun := false
+
+	for i := range len(line) {
+		if isInlineWhitespace(line[i]) {
+			inRun = true
+
+			continue
+		}
+
+		if inRun {
+			sb.WriteByte(' ')
+
+			inRun = false
+		}
+
+		sb.WriteByte(line[i])
+	}
+
+	if inRun {
+		sb.WriteByte(' ')
+	}
+
+	return sb.String()
+}
+
 // GetConditionLineNumber returns the index of the line declaring conditionName, or
 // -1. `less` does not match a declaration of `less_than`.
 func GetConditionLineNumber(conditionName string, lines []string) int {
-	return declarationIndex(lines, "condition "+conditionName, keepSpaces)
+	return declarationIndex(lines, "condition "+conditionName)
 }
 
 // GetTypeLineNumber returns the index of the line declaring typeName, or -1.
 // `document` does not match a declaration of `documentation`.
 func GetTypeLineNumber(typeName string, lines []string) int {
-	return declarationIndex(lines, "type "+typeName, keepSpaces)
+	return declarationIndex(lines, "type "+typeName)
 }
 
 // GetExtendedTypeLineNumber returns the index of the line extending typeName, or -1.
 func GetExtendedTypeLineNumber(typeName string, lines []string) int {
-	return declarationIndex(lines, "extend type "+typeName, keepSpaces)
+	return declarationIndex(lines, "extend type "+typeName)
 }
 
 // GetRelationLineNumber returns the index of the line defining relation, or -1.
 // `owner` does not match a definition of `owner_group`.
 func GetRelationLineNumber(relation string, lines []string) int {
-	return declarationIndex(lines, "define "+relation, normalizeSpaces)
+	return declarationIndex(lines, "define "+relation)
 }
 
 type StartEnd struct {
--- a/pkg/go/utils/line-numbers.go
+++ b/pkg/go/utils/line-numbers.go
@@ -28,7 +28,8 @@ func IsNameByte(b byte) bool {
 	return nameBytes[b]
 }
 
-// name ends there, or -1. Prefix always spans whole words (`type <name>`, `define
+// declarationIndex returns the index of the first line beginning with prefix whose
+// name ends there, or -1. prefix always spans whole words (`type <name>`, `define
 // <name>`, `condition <name>`), so requiring the next byte to end the name is what
 // stops `document` matching a declaration of `documentation`.
 func declarationIndex(lines []string, prefix string) int {
--- a/pkg/go/utils/line-numbers.go
+++ b/pkg/go/utils/line-numbers.go
@@ -5,21 +5,27 @@ import (
 	"strings"
 )
 
-// IsNameByte reports whether b can appear inside a declaration name. IDENTIFIER
-// admits letters, digits, `_` and MINUS, and EXTENDED_IDENTIFIER adds SLASH and DOT
+// nameBytes marks every byte that can appear inside a declaration name, keyed by
+// the bytes themselves so each entry is visibly correct. IDENTIFIER admits letters,
+// digits, `_` and MINUS, and EXTENDED_IDENTIFIER adds SLASH and DOT
 // (OpenFGALexer.g4), so `core.doc` is one name rather than `core` and a terminator.
+var nameBytes = [256]bool{
+	'a': true, 'b': true, 'c': true, 'd': true, 'e': true, 'f': true, 'g': true,
+	'h': true, 'i': true, 'j': true, 'k': true, 'l': true, 'm': true, 'n': true,
+	'o': true, 'p': true, 'q': true, 'r': true, 's': true, 't': true, 'u': true,
+	'v': true, 'w': true, 'x': true, 'y': true, 'z': true,
+	'A': true, 'B': true, 'C': true, 'D': true, 'E': true, 'F': true, 'G': true,
+	'H': true, 'I': true, 'J': true, 'K': true, 'L': true, 'M': true, 'N': true,
+	'O': true, 'P': true, 'Q': true, 'R': true, 'S': true, 'T': true, 'U': true,
+	'V': true, 'W': true, 'X': true, 'Y': true, 'Z': true,
+	'0': true, '1': true, '2': true, '3': true, '4': true,
+	'5': true, '6': true, '7': true, '8': true, '9': true,
+	'_': true, '-': true, '/': true, '.': true,
+}
+
+// IsNameByte reports whether b can appear inside a declaration name.
 func IsNameByte(b byte) bool {
-	switch {
-	case b >= 'a' && b <= 'z', b >= 'A' && b <= 'Z', b >= '0' && b <= '9':
-		return true
-	}
-
-	switch b {
-	case '_', '-', '/', '.':
-		return true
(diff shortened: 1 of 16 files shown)


pkg/go/utils/line-numbers.go
```diff
@@ -5,27 +5,87 @@ import (
...
+// endOrComment reports whether the remainder of a declaration line is empty or is
+// only a trailing comment, mirroring the reference's `^(\s+#.*)?$`. The `#` must be
+// preceded by whitespace so a `#` glued to the name isn't treated as a comment —
+// `type doc#x` is not a declaration of `doc`.
+func endOrComment(rest string) bool {
+	if rest == "" {
+		return true
+	}
+
+	trimmed := strings.TrimLeft(rest, " \t")
```
Siddhant-K-code: `WHITESPACE` also permits `\f`, but this only trims spaces/tabs. Include `\f` here (and in `startsWith`) so valid declarations don’t fall back to a `0:0` location; add a regression test.

It also applies on L42

Siddhant-K-code (dismissed):

senojj:
@SoulPancake why force the caller to validate the expected tail of the string? You could simply get the index immediately following the provided prefix -- if one exists -- and check for a word termination character.

It seems that an invariant of this function is that `prefix` be a full word or words, and not just a partial word. Therefore, it makes more sense to assert that any matching prefix be followed by a word terminator. There is no need for the `rest` argument.

SoulPancake:
Shared fixtures causing the npm audit failure to show up in this PR, checking that.

Edit: Letting that remain as a separate fix PR to not digress from the topic here.

Siddhant-K-code (commented):
Tests/lint seems to be failing - can i send a commit to fix it or you wanna address it, @SoulPancake ?

SoulPancake:
@Siddhant-K-code 
https://github.com/openfga/language/pull/652#issuecomment-5201162786

Siddhant-K-code (dismissed):
Thanks for addressing the review comments! Changes look good to me; the prefix matching fix is clear & well tested. % address the lint failing please before merge!

I'd also wait for review from @senojj

pkg/go/utils/line-numbers.go
```diff
@@ -5,28 +5,76 @@ import (
...
+	case b >= 'a' && b <= 'z', b >= 'A' && b <= 'Z', b >= '0' && b <= '9':
+		return true
+	}
+
+	switch b {
+	case '_', '-', '/', '.':
+		return true
+	}
+
+	return false
```
senojj: This can be done much more efficiently with a lookup table:
```go
var nameTable = [256]byte{
	//   1    2    3    4    5    6    7    8    9    A    B    C    D    E    F
	0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, // 0
	0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, // 1
	0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, '-', '.', 0x0, // 2
	'0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, // 3
	0x0, 'A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K', 'L', 'M', 'N', 'O', // 4
	'P', 'Q', 'R', 'S', 'T', 'U', 'V', 'W', 'X', 'Y', 'Z', 0x0, '/', 0x0, 0x0, '_', // 5
	0x0, 'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j', 'k', 'l', 'm', 'n', 'o', // 6
	'p', 'q', 'r', 's', 't', 'u', 'v', 'w', 'x', 'y', 'z', 0x0, 0x0, 0x0, 0x0, 0x0, // 7
	0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, // 8
	0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, // 9
	0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, // A
	0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, // B
	0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, // C
	0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, // D
	0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, // E
	0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, // F
}

func IsNameByte(b byte) bool {
    return nameTable[b] != 0x0
}
```

pkg/go/utils/line-numbers.go
```diff
@@ -5,28 +5,76 @@ import (
...
+}
+
+// declarationIndex returns the index of the first line beginning with prefix whose
+// name ends there, or -1. prefix always spans whole words (`type <name>`, `define
+// <name>`, `condition <name>`), so requiring the next byte to end the name is what
+// stops `document` matching a declaration of `documentation`.
+func declarationIndex(lines []string, prefix string, normalize func(string) string) int {
 	return slices.IndexFunc(lines, func(line string) bool {
-		return strings.HasPrefix(strings.TrimSpace(line), "condition "+conditionName)
+		trimmed := normalize(strings.TrimSpace(line))
```
senojj: why is the normalize function necessary here?

pkg/go/utils/line-numbers.go
SoulPancake: Good idea!

One small thing: the posted table swaps `/` (0x2F) and `\` (0x5C), which would make `type internal` match `type internal/doc` again (`/` is valid in `EXTENDED_IDENTIFIER`). Using a keyed `[256]bool` literal instead

SoulPancake: In whcih case we can also skip the compare

pkg/go/utils/line-numbers.go
SoulPancake: It's matching the JS reference really 
relation lines get their spaces collapsed there, so `define  owner: `still resolves:
https://github.com/openfga/language/blob/75569ab6598080cc8899c37d5505187fe60b9d4c/pkg/js/util/line-numbers.ts#L44-L48
Dropping it would have Go and JS report different error locations for the same model, ran a case locally to confirm

Siddhant-K-code (commented):
@SoulPancake, JS tests/lint seems to be failing

SoulPancake:
> @SoulPancake, JS tests/lint seems to be failing

https://github.com/openfga/language/pull/652#issuecomment-5201162786

pkg/go/utils/line-numbers.go
```diff
@@ -5,28 +5,82 @@ import (
...
+// spaces, matching the reference, where ` {2,}` normalization is applied in
+// getRelationLineNumber alone.
+func keepSpaces(line string) string { return line }
+
+// normalizeSpaces collapses runs of spaces into one, mirroring the reference's
+// ` {2,}` normalization, so `define  owner:` is matched the same as `define owner:`.
+func normalizeSpaces(line string) string {
+	for strings.Contains(line, "  ") {
+		line = strings.ReplaceAll(line, "  ", " ")
+	}
```
senojj: I notice that this only handles space characters (hexadecimal "20"). What happens when tabs, nbsp, and other whitespace characters are present?

SoulPancake: nbsp can't reach here — [`WHITESPACE`](https://github.com/openfga/language/blob/e81c90728812a5bbad1c02bdc172c6ca625c9550/OpenFGALexer.g4#L107) is `'\t' | ' ' | '\u000C'` only, so `define<nbsp>owner:` dies in the lexer first:

```
syntax error at line=5, column=10: token recognition error at: ' '
```

Tabs are a miss. [`WHITESPACE` is one token matching a run](https://github.com/openfga/language/blob/e81c90728812a5bbad1c02bdc172c6ca625c9550/OpenFGAParser.g4#L13-L16), so `define<tab>owner:` is valid DSL, and the helper returns -1 — the error lands on line 0:

```
tab sep   transformation error at line=0, column=0:  relation owner already exists on type document
2sp sep   transformation error at line=4, column=12: relation owner already exists on type document
```

The reference has the same thing, though — [`getRelationLineNumber`](https://github.com/openfga/language/blob/e81c90728812a5bbad1c02bdc172c6ca625c9550/pkg/js/validator/validate-dsl.ts#L339) only folds ` {2,}`, and [`getTypeLineNumber`/`geConditionLineNumber`](https://github.com/openfga/language/blob/e81c90728812a5bbad1c02bdc172c6ca625c9550/pkg/js/validator/validate-dsl.ts#L294-L327) don't fold at all, so a tab misses there too 🤔 

Might be worth fixing it in both while we're at it? Collapsing runs of `[ \t]` instead of just spaces, and doing it in all the helpers rather than only the relation one. It'd have to go into `validate-dsl.ts` at the same time, otherwise the shared corpus stops catching drift between the two.

SoulPancake: addressed

PR #653, opened 2026-08-05
feat(pkg/go/graph): attribute weighted-graph nodes to their relation

SoulPancake:
Edges already carry a relationDefinition (#494), but nodes do not, so a failure detected on an operator or logical node cannot name the relation it belongs to. That attribution is what lets a graph-derived finding be resolved to a line and column, so it is a prerequisite for reporting weighted-graph errors to users.

Add relationDefinition to WeightedAuthorizationModelNode with a GetRelationDefinition() accessor mirroring the one already on edges, and populate it in the builder:

- SpecificType and SpecificTypeWildcard nodes get the empty string; a type or wildcard is reachable from many relations, so it belongs to none.
- SpecificTypeAndRelation nodes are their own relation definition.
- Operator and logical grouping nodes get the relation whose rewrite created them. The value is built from typeDef and relation, not from the enclosing node, so nested operators attribute to the relation rather than to the operator above them.

Also adds SplitRelationDefinition, which splits "objectType#relation" into its parts. This is the helper that exists unexported as getRelationAndObject on feat/language-new-version; it is renamed here so the name matches the order of the return values, and it treats an empty relation definition the same as a malformed one so callers can handle "no attribution" in one branch.

AddNode and GetOrAddNode take an extra parameter, which is a breaking change to two exported functions. The core server only reads nodes and never constructs them, so it is unaffected; verified by building openfga/openfga against this tree.

Behaviour is otherwise unchanged: no weights, edges, labels or errors differ, and the existing suite passes untouched apart from the added argument at the call sites.

## Review Checklist


senojj:
@SoulPancake if this is to assist with error handling during graph traversal, why is the edge not in scope at the time of error?

SoulPancake:
@senojj 

The edge isn't in scope at those sites — the weight functions only receive a node ID:

```go
func (wg *WeightedAuthorizationModelGraph) calculateNodeWeightWithMaxStrategy(nodeID string) error {
	node := wg.nodes[nodeID]
	edges := wg.edges[nodeID]

	if len(edges) == 0 && node.nodeType != SpecificType && node.nodeType != SpecificTypeWildcard {
		return fmt.Errorf("%w: %s node does not have any terminal type to reach to", ErrInvalidModel, node.uniqueLabel)
	}
```

Same for the mixed and enforce-type strategies. `ancestorPath` exists up in `calculateNodeWeight` but stops at `calculateNodeWeightFromTheEdges` and never reaches these functions, and since `edges` is keyed by source there's no way to ask what points *at* a node. So the node is all we have to name the failure with.

It matters when the label is synthesised: `doc#ttu:parent#member` names the tupleset it reads through, never `viewer`, which is the relation the user would actually edit.

Carried over from Yissell's `feat/language-new-version`, where it's consumed at [five of these same sites](https://github.com/openfga/language/blob/9583a0b43d620ad6c226c8341ac5dc43386282ec/pkg/go/graph/weighted_graph.go#L496) to return a structured error instead of a raw label.

That said, I checked and the incoming edge does carry the same `relationDefinition` as the node, so threading an edge or `ancestorPath` down would work too. What would you recommend?

senojj:
@SoulPancake in graph theory, the "degree" of a vertex is equal to the number of edges incident on it. In a directed graph, the degree is split into two components, the "indegree" and the "outdegree". 

The "indegree" is what we are interested in here, and represents the number of incoming edges incident on the vertex. The key property to observe here is that the indegree of a vertex may be greater than 1. 

In a scenario where the vertex's indegree is equal to 4, what would the relation definition be equal to?

SoulPancake:
@senojj 

Thanks for the patience on these reviews.

 for operator and direct-grouping nodes: their labels carry the relation (`type#relation:operator:index`, `type#direct:relation`), so they're never shared across relations. The TTU grouping label doesn't, though — `type#ttu:<tupleset>#<computed>` so every relation rewriting through the same tupleset and computed relation lands on one node. It needs the TTU to sit under an operator or grouping node with a multi-type tupleset, i.e. the `parentNode.nodeType != SpecificTypeAndRelation && len(directlyRelated) > 1` guard in `parseTupleToUserset`; a bare TTU directly under a relation builds no grouping node at all. Four relations doing `member from parent`:

```
doc#ttu:parent#member   indegree=4   node.relationDefinition="doc#a_rel"
                        predecessors: doc#a_rel:union:0 ... doc#d_rel:union:0
                        edge relDefs: {doc#a_rel, doc#b_rel, doc#c_rel, doc#d_rel}
```

I constructed this to check your indegree-4 question, and the four predecessors turn out to be distinct operator nodes rather than parallel edges from one source. The stored value is whichever relation sorts first, since `Build` sorts alphabetically and `GetOrAddNode` keeps the first writer, so three of the four relations are attributed to a relation they don't own. Nothing consumes the field yet — the weight errors still print `node.uniqueLabel` — so no output is wrong today. It's the stored value that's wrong, and it would start showing up in error messages as soon as something reads it.

What gives a node one correct value is having the relation in its label, not its indegree — `doc#viewer` can have indegree 10 and still be right, because only `viewer` can ever write to it. So the invariant is one relation per node, not one edge. It reproduces from a clean checkout, but please correct me if there's a shape here I'm not seeing ? and let me know how you'd like the TTU case handled 🤔

senojj:
Yes, this is proof that a logical TTU edge can have an indegree value greater than 1.

Given the following model:
```fga
model
  schema 1.1

type user

type org
  relations
    define member: [user]

type team
  relations
    define member: [user]

type document
  relations
    define parent: [org, team]
    define viewer: rel_a or rel_b or rel_c or rel_d
    define rel_a: [user] or member from parent
    define rel_b: [user] or member from parent
    define rel_c: [user] or member from parent
    define rel_d: [user] or member from parent
```

Here is a visualization:
```text
document#viewer
└──[2][R]── document#viewer:union:0
            ├──[2][R]── document#rel_a
            │           └──[2][R]── document#rel_a:union:0
            │                       ├──[1][D]── user
            │                       └──[2][L]── document#ttu:parent#member
            │                                   ├──[2][T]── document#parent ↦ org#member
            │                                   │           └──[1][D]── user
            │                                   └──[2][T]── document#parent ↦ team#member
            │                                               └──[1][D]── user
            ├──[2][R]── document#rel_b
            │           └──[2][R]── document#rel_b:union:0
            │                       ├──[1][D]── user
            │                       └──[2][L]── document#ttu:parent#member
            │                                   ├──[2][T]── document#parent ↦ org#member
            │                                   │           └──[1][D]── user
            │                                   └──[2][T]── document#parent ↦ team#member
            │                                               └──[1][D]── user
            ├──[2][R]── document#rel_c
            │           └──[2][R]── document#rel_c:union:0
            │                       ├──[1][D]── user
            │                       └──[2][L]── document#ttu:parent#member
            │                                   ├──[2][T]── document#parent ↦ org#member
            │                                   │           └──[1][D]── user
            │                                   └──[2][T]── document#parent ↦ team#member
            │                                               └──[1][D]── user
            └──[2][R]── document#rel_d
                        └──[2][R]── document#rel_d:union:0
                                    ├──[1][D]── user
                                    └──[2][L]── document#ttu:parent#member
                                                ├──[2][T]── document#parent ↦ org#member
                                                │           └──[1][D]── user
                                                └──[2][T]── document#parent ↦ team#member
                                                            └──[1][D]── user
```
It can be seen that vertex `document#ttu:parent#member` has an indegree of 4, with each edge coming from a distinct specific type and relation.

yissellokta:
we are not making any change to the weighted graph, any change will need to be vetted before

yissellokta (changes requested):
Avoid changes into the weighted graph, validate any change in the weighted graph with me or Joshua

PR #660, opened 2026-08-12
refactor(pkg/go)!: restructure validation and errors

SoulPancake:
## What this does

Restructures Go validation so findings are constructed at the point of failure and returned up the stack, replacing the collector threaded down through every validator.

- `Finding` is one diagnostic; its fields are the wire shape (`msg`, `line`, `column`, `file`, `metadata{…}`).
- `Findings []*Finding` is the collection and, when non-empty, the error — the `go/scanner.ErrorList` pattern. `ValidateDSL`/`ValidateJSON` return `nil` or a `Findings`.
- `Kind` (the wire `errorType`) is the single identity axis.
- Leaf validators take only what they check and return `*Finding`; the calling phase stamps position and file/module, since it is what holds the source lines and the proto `SourceInfo`.
- Later structural phases are skipped once anything is found.

## Removed

The severity/category/cause classification and its table, the sentinels and scoped error types (`pkg/go/errors` returns to its pre-branch file), the collector, and the engine/options/summary/report types. Severity was one uniform value across every code and nothing consumed the rest; each removed piece can return additively when something emits or consumes it.

## Output compatibility

Both shared corpora pass with byte-identical message, `errorType`, symbol, line and column. Line lookups now fold runs of inline whitespace (space, tab, form feed — the lexer's `WHITESPACE` alphabet) so tab-separated declarations resolve; columns still resolve against the raw line.

## Breaking

The exported surface is now `Finding`, `Findings`, `Kind` (+ constants), `Range`, `Metadata`, `ValidateDSL`, `ValidateJSON`.


pkg/go/validation/error_collector.go
```diff
@@ -132,27 +204,102 @@ func (c *ErrorCollector) addError(message string, errorType ValidationErrorType,
...
+		return "", "", ""
+	}
+}
+
 // RaiseInvalidName raises an invalid name error.
 func (c *ErrorCollector) RaiseInvalidName(symbol, clause string, typeName *string, lineIndex *int, meta *Meta) {
 	var message string
+	// A nil typeName means the offending name is a type rather than a relation on
+	// one, which changes both the message and the scope of the finding.
+	errorScope := scope{objectType: symbol, category: fgaerrors.ErrorKindObjectType}
```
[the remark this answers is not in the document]
SoulPancake: Resolved in dc25ffc

docs/validation/model/invalid-schema.md
```diff
@@ -6,149 +6,68 @@
...
 
-### Malformed schema structure:
 ```
 model
-  # Error: Schema declaration without version
-  schema
+  schema 0.9
 
 type user
 ```
```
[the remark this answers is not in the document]
SoulPancake: I think since there's no such linter and it's used nowhere else we shouldn't worry about it but we can open an issue for a markdown linter to be present for this repo generally

pkg/go/errors/model_error_kind.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	return values
+}()
+
+// String returns the wire name, or a diagnostic form for a value with none.
+func (m ModelErrorKind) String() string {
+	if name, ok := modelErrorKindNames[m]; ok {
+		return name
+	}
```
senojj: Unsure of the goal here, but a `switch` statement is almost always going to be more efficient than a `map` lookup.

So I think a `switch` statement here is preferable, and then one in a function like `ModelErrorKindFromName`. `MarshalText` should use `(ModelErrorKind).String()` and `UnmarshalText` should use `ModelErrorKindFromName`.

pkg/go/validation/error_collector.go
```diff
@@ -66,30 +69,78 @@ func NewErrorCollector(lines []string) *ErrorCollector {
...
 	lineIndex *int, meta *Meta, customResolver ErrorCustomResolver) {
-	var line *LineRange
-	var column *ColumnRange
+	c.addScopedError(message, errorType, symbol, lineIndex, meta, customResolver, scope{})
+}
+
+// addScopedError adds an error that knows which type, relation or condition it
+// concerns. Callers with nothing to add beyond the symbol use addError instead.
+func (c *ErrorCollector) addScopedError(message string, errorType ValidationErrorType, symbol string,
+	lineIndex *int, meta *Meta, customResolver ErrorCustomResolver, errorScope scope) {
```
senojj: I'm not sure I understand this function, the caller has all of the context necessary to construct the relevant error value. Why have the caller encode that context into this function's more generalized context?

pkg/go/validation/error_info.go
```diff
@@ -0,0 +1,243 @@
...
+	InvalidSchema,
+	InvalidSyntax,
+	TypeRestrictionCannotHaveWildcardAndRelation,
+	ConditionNotDefined,
+	ConditionNotUsed,
+	DifferentNestedConditionName,
+	MultipleModulesInFile,
+	CyclicRelation,
+	InvalidSchemaVersion,
+}
```
senojj: constructs used only by tests should remain with the tests. that way they don't get compiled in the release binary, or included in vendored code.

pkg/go/validation/error_collector.go
SoulPancake: The caller can't work out the position (needs the DSL source the collector holds) or the severity and sentinel (from `errorInfoByType`, keyed on the code). Metadata is read back off the scope so it can't disagree with `errors.As`.

You're right about the scope part though. Raise sites passed loose strings and this picked the error type off a category table; e5663ec has them build the `ModelError` directly instead.

`symbol`, `lineIndex` and `customResolver` are just position inputs but I think we can collapse them in one

pkg/go/errors/model_error_kind.go
```diff
@@ -0,0 +1,123 @@
...
+	case ErrorKindCondition:
+		return "condition"
+	case ErrorKindInvalidModel:
+		return "invalid-model"
+	default:
+		// ModelErrorKindUnspecified lands here too: the zero value has no name by
+		// design, so it marshals as a failure rather than as a category.
+		return ""
+	}
+}
```
senojj: This function serves very little utility; its contents should go directly in the `String()` method.

pkg/go/errors/model_error_kind.go
```diff
@@ -0,0 +1,123 @@
...
+		return ""
+	}
+
+	return fmt.Sprintf("ModelErrorKind(%d)", int(m))
+}
+
+// IsValid reports whether m is a declared category with a wire name.
+func (m ModelErrorKind) IsValid() bool {
+	return m.wireName() != ""
+}
```
senojj: The `isValid` function should not rely on generating a string value to determine validity. `ModelErrorKind` is a closed enum of only a handful of values; just use a switch statement here.

An enum variant's string value is orthogonal to the validity of the variant.

SoulPancake:
@senojj addressed your comments, feel free to re-review 😄

pkg/go/validation/source.go
```diff
@@ -0,0 +1,266 @@
...
+}
+
+// foldInline collapses every run of inline whitespace into one space, so
+// `define\towner:` is matched like `define owner:`. Only space, tab and form
+// feed fold — exactly the lexer's WHITESPACE alphabet (OpenFGALexer.g4);
+// anything else fails to lex and can never reach a line lookup. Lines already
+// normal are returned unchanged. Matching only: columns are still resolved
+// against the raw line, so folding never shifts a reported position.
+//
+// TODO(SoulPancake): replace with utils.NormalizeWhitespace once #652 merges,
```
SoulPancake, line 26: https://github.com/openfga/language/pull/652

PR #663, opened 2026-08-18
fix(js): refresh npm audit remediation in #651

Siddhant-K-code:
## Description

Updates the source branch for #651 with the current audit remediation:

- merges current `main`, including the existing `fast-uri` 3.1.5 fix
- resolves the lockfile conflict using `main`'s patched dependency resolutions
- raises the `js-yaml` override from `^4.3.0` to `^4.3.1`, the first version outside the latest advisory range

Merging this PR updates #651 directly. A direct push to the upstream source branch was not permitted for the current contributor credentials.

## Testing

Using Node 24.19.0 and npm 11.17.0:

- `npm ci`
- `npm audit` — 0 vulnerabilities
- `npm run lint`
- `npm run format:check`
- `npm run typecheck`
- `npm run build:light`
- `npm test -- --runInBand` — 666 passed, 27 skipped

### DIFF
--- a/.github/workflows/pkg-java-build.yaml
+++ b/.github/workflows/pkg-java-build.yaml
@@ -26,7 +26,7 @@ jobs:
         uses: docker/setup-buildx-action@bb05f3f5519dd87d3ba754cc423b652a5edd6d2c # v4.2.0
 
       - name: Set up JDK
-        uses: actions/setup-java@03ad4de0992f5dab5e18fcb136590ce7c4a0ac95 # v5.6.0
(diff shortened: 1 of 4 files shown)


SoulPancake (approved):

SoulPancake:
Might need to rebase and amend author
Although I believe we should also add codex to EasyCLA thing like we do for Copilot

Siddhant-K-code:
> Although I believe we should also add codex to EasyCLA thing like we do for Copilot

We need to raise request to CNCF about that!

PR #664, opened 2026-08-18
feat: support $expression reserved condition name on type restrictions

yissellokta:
Adds `$expression` as a reserved **inline-expression** condition name usable on type
  restrictions, e.g.:

  type document
    relations
      define viewer: [user, user with cond1, user with $expression]

  `$expression` behaves like a named condition on the wire (carried on
  `RelationReference.Condition`) but is **not** declared in the model's `conditions`
  block — the CEL body is supplied per-tuple at write time (handled downstream in the
  OpenFGA server). This is the language-layer groundwork for FGA Dynamic Conditions.

  It is valid **only** inside a type restriction — declaring `condition $expression(...)`
  remains illegal.

  Works on direct, userset (`group#member with $expression`), and TTU
  (`parent: [group with $expression]`) restrictions.

  ## Changes

  **Grammar**
  - Lexer: add `DOLLAR_EXPRESSION` token before `IDENTIFIER` (additive; `$` was
    otherwise unused, so no collision).
  - Parser: accept it inline in `relationDefTypeRestriction` as
    `(conditionName | DOLLAR_EXPRESSION)` rather than widening `conditionName`, so it
    stays illegal as a condition declaration.
  - Regenerated Go / JS / Java parsers via `make antlr-gen`.

  **Transformers (Go / JS / Java)**
  - Set `Condition = "$expression"` when the token is present. `jsontodsl` needs no
    change in any binding — it already emits any non-empty condition verbatim, so models
    round-trip DSL → JSON → DSL unchanged.

  **JS validator**
  - Exempt the reserved name from the "condition must be declared" check (would
    otherwise reject a valid single-`$expression` model). The existing condition-aware
    duplicate-restriction check already rejects two `$expression` on the same terminal
    type (e.g. `[user with $expression, user, user with $expression]`), while two
    *different* named conditions stay legal.

  **Weighted graph**
  - No builder change: `$expression` flows through as an opaque condition string on
    direct, userset, and TTU edges.

  ## Tests

  - Go: transformer parse + JSON round-trip + reject-as-declaration
    (`pkg/go/transformer/inline_expression_test.go`); weighted-graph flow-through on
    direct/userset/TTU edges (`pkg/go/graph/inline_expression_test.go`).
  - JS: transformer + validator, incl. duplicate-rejection and declaration-rejection
    (`pkg/js/tests/inline-expression.test.ts`).
  - Shared: two valid semantic cases added to `tests/data/dsl-semantic-validation-cases.yaml`.

  Full Go and JS suites pass.

  ## Notes

  - Purely additive — no breaking change. A pre-existing model using the literal
    `$expression` as a named condition can't exist (`$` was never a valid lexer
    character).
  - The Go transformer performs no semantic dedup, so the "no two `$expression` per
    terminal type" rule is enforced in the JS/Java validators here; Go enforcement lands
    in the OpenFGA server's typesystem in a follow-up.
  - Java was not compiled locally (Maven Central unreachable in my env); the regenerated
    Java parser exposes the `DOLLAR_EXPRESSION()` accessor the listener edit uses — please
    rely on CI to compile-verify.

## Summary by CodeRabbit

* **New Features**
  * Added support for inline `$expression` conditions in relation type restrictions.
  * Inline expressions are now recognized and preserved across supported language integrations.
* **Validation**
  * Reserved `$expression` conditions are accepted without requiring a separately named condition.
  * Added validation coverage for inline expressions on standard and userset restrictions.


senojj (approved):

yissellokta:
@claude review

